PR #41, opened 2020-06-19
Add implicit import wrapper

jacksteamdev:
This PR adds two new dynamic import wrappers and refactors the old one.

### Content Scripts

Content scripts now have their own import wrapper. It simply uses dynamic import to load the content script module and does nothing with events, since content scripts are always persistent.

### Background Scripts

The new default import wrapper is the implicit import wrapper. It walks the `chrome` object and wraps any events that it finds. If your extension uses many permissions, the `chrome` object will have more namespaces, and it may take slightly more time (~15ms) to load the background page. Consider the following if 15ms is too much for you.

The old import wrapper is now the explicit import wrapper, and uses the same API as before. It is quite fast (~5ms), but wake events that are not defined in `dynamicImportWrapper.wakeEvents` will be lost. Symptoms of this would be things like having to click the browser action twice to register `browserAction.onClicked`, or missed events from `alarms.onAlarm`.

### Motivation

These wrappers are necessary because Rollup only supports ES6 modules when `output.dir` is used, which is necessary for this plugin to work the way that it does. Code splitting is also non-optional with `output.dir`. This means that we need all script environments to support ES6 modules.

However, in Chrome, background and content scripts do not support ES6 modules! They do support the dynamic import function, which allows us to async load modules into a script.

This introduces another complexity with non-persistent background pages. Events that wake the background page are fired at the end of first event loop, before the dynamic import completes. These import wrappers capture wake events and dispatch them after the dynamic import resolves.

### Firefox incompatibility

Please note that none of these wrappers are compatible with the Firefox browser. Firefox does not currently support dynamic imports in web extensions. Further development will be necessary to add Firefox build support.

While we can load code into a background script through the dom using a script tag in Firefox, this is not an option in content scripts. We will need to build each content script on its own.

Meanwhile, a workaround to obtain a Firefox build is to pass the Chrome extension build through Parcel. It will be a considerably larger bundle, but it does work.


fmonper1:
Can you expand a bit more on the Firefox build with Parcel? 

Good work with the plugin!

jacksteamdev:
Thanks! I've used Parcel and [parcel-plugin-web-extension](https://github.com/kevincharm/parcel-plugin-web-extension) to create Firefox compatible builds in the past. 

It's nice b/c it's a zero-config option that can work from the same project. Just install Parcel and the plugin, point it to the build or the src folder and everything works.

The build is much bigger and webpack-y, but it works in a pinch.

jessuni:
Great work! It'd be nice if it supports Firefox build as well in the future. We go for rollup because of its smaller bundle size.

PR #63, opened 2021-02-16
fix: jsx & tsx not converted

ZakaryCode:
When assets is generated, jsx & tsx file path not converted

Community guidelines:

      and

This PR contains:

- [x] bugfix

Are tests included?

Breaking Changes?

- [x] no

### Description

There should be a better solution. This solution is based on the current code.

### DIFF
--- a/src/manifest-input/index.ts
+++ b/src/manifest-input/index.ts
@@ -403,7 +403,7 @@ export function manifestInput(
                 ? rest
                 : {
                     js: js
-                      .map((p) => p.replace(/\.ts$/, '.js'))
+                      .map((p) => p.replace(/\.[jt]sx?/g, '.js'))
                       .map(memoizedEmitter),
                     ...rest,
                   }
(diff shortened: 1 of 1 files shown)


jacksteamdev (changes requested):
@ZakaryCode Thanks for the PR!

Can you add some tests to make sure that this is working the way we want?

The tests need to check two things:

- `some-file.tsx` is replaced with `some-file.js`
- `some.tsx.js` is not changed

I know `some.tsx.js` is a strange file name, but we want to make sure we don't break things. :wink:

ZakaryCode:
> @ZakaryCode Thanks for the PR!
> 
> Can you add some tests to make sure that this is working the way we want?
> 
> The tests need to check two things:
> 
> * `some-file.tsx` is replaced with `some-file.js`
> * `some.tsx.js` is not changed
> 
> I know `some.tsx.js` is a strange file name, but we want to make sure we don't break things. 😉

Sorry, I just saw it, but I don't think `some.tsx.js` needs to be fixed. In principle, this is the normal introduction of it.

ZakaryCode:
FYI, that we writed `web_accessible_resources` that it was incorrect.

```json
  // script not import
  "web_accessible_resources": [
    "./src/content.js",
  ],
  // script import
  "web_accessible_resources": [
    "src/content.js",
  ],

```

Sorry, I didn't open a new issue to explain this, but although these two issues are not the same problem, they both lead to `content_scripts` are not working properly.

jacksteamdev:
@ZakaryCode I wrote tests for this feature and put you as co-author for #67

Thanks for contributing!

Issue #64, opened 2021-02-17
TSX content script file extension makes [name]-[hash].tsx output file in dist


joegilley:
### Google search terms

I tried latest `@rollup/plugin-typescript` and `rollup-plugin-typescript2` in case it was the typescript plugin.

### Describe the bug

Specifying a TSX file in `content_scripts` creates a file in `dist/assets` with a TSX extension instead of JS.  The contents of the file don't need to actually have React, just the `.tsx` extension is enough to trigger the issue.

### How do we reproduce?

Change the line [here](https://github.com/extend-chrome/ts-react-boilerplate/blob/3d5217985198723411e085f6bbac5ee2eba275c8/src/manifest.json#L5) to end in `.tsx` and rename the associated file to end with `.tsx`, and build w/ `rollup -c`.

### Expected behavior

No typescript files in the `dist` folder

### Actual behavior

![image](https://user-images.githubusercontent.com/1476969/108151102-d407a280-708a-11eb-864a-ffadf77e18e2.png)

The contents of the file seem off as well:
```js
(function () {
	'use strict';

	const importPath = /*@__PURE__*/JSON.parse('"../content/index.tsx"'); // 👈 extension here is wrong i think

	import(importPath);

}());
```

Note, though, the expected file in `dist/content/index.js` is correct. ¯\\_(ツ)_/¯ 

### Screenshots
output:
![image](https://user-images.githubusercontent.com/1476969/108150629-d1f11400-7089-11eb-8fdf-d33df400462d.png)
no build errors:
![image](https://user-images.githubusercontent.com/1476969/108150648-dc131280-7089-11eb-805e-7000c94a1c83.png)

### Please complete the following information:

- Library Version: 3.5.3
- Operating System: Mac OS Big Sur
- Browser: Chrome 
- Node Version: v14.15.4

### Additional context

P.S. thank you for this library, it is just so so much better than anything else I've found, primarily the same 3 year old tutorial being reposted over and over.  The live reloading alone is worth gold.

joegilley:
#63 related

jacksteamdev:
@joegilley Thanks for putting up a PR! I put you as co-author on #67, which fixes this issue.

BTW, I've refactored the tests to make it easier for contributors to add tests. Now we just need to duplicate `__fixtures__/extensions/basic-js/` and `__tests__/e2e__basic-js.test.ts` and update the tests appropriately.

PR #65, opened 2021-02-17
format tsx content_scripts as js

joegilley:
a quickie fix for #64 .  Sorry, I dont know how to add a test, but I used it in my project and it worked perfectly.


joegilley:
can't believe i didnt see #63 when debugging this 🤦

Issue #56, opened 2020-10-21
Support firefox / cross browser compatibility


anarchylimes:
### Is your feature request related to a problem? Please describe.

No issue per-say but this rollup plugin would benefit people greatly if it worked in firefox as well.

### Describe the solution you'd like

To be able to develop cross-browser extensions using only this plugin. There is a fork already for webextension-polyfill, which is the basis of cross-browser extension development

### Describe alternatives you've considered

I've already done a rollup configuration that allows development of firefox plugins but only by importing every file separately. This would make it so much faster to create browser extensions

jacksteamdev:
A relevant comment by @avi12 from [another issue](https://github.com/extend-chrome/rollup-plugin-chrome-extension/issues/60#issuecomment-752077425):

> But, for a strange reason, not on Firefox (loaded via web-ext)
> Example repo: https://github.com/avi12/steam-id-finder/tree/main

jacksteamdev:
The issue here is that Firefox doesn't support dynamic imports in web extensions, which we use to support Rollup's code splitting feature. What we need to do is bundle each content script and background script as an IIFE. This basically results in a mixed format build, which Rollup doesn't support natively.

I've implemented the core for this feature already with `options.iifeJsonPaths`, which uses the ESM bundle as a virtual file system to build the other formats. The plan is to export an [output plugin](https://rollupjs.org/guide/en/#using-output-plugins) called `FirefoxAddon` which can be used to create seperate builds for both targets.

avi12:
I tried importing functions using an `import` statement, e.g.
```js
import { someFunction } from "./some-script";
```
but it didn't work, despite not being a dynamic import (i.e. `improt("./some-script.js")`)

Glench:
any development on this? I could really use it while developing [ExtPay](https://github.com/glench/ExtPay).

avi12:
@Glench What I used as a "workaround" is [this code](https://github.com/NekitCorp/chrome-extension-svelte-typescript-boilerplate)

Glench:
@avi12 that's actually what I tried using but it's failing with a bunch of errors in firefox

avi12:
[...]

PR #75, opened 2021-05-21
Don't call `Object.keys` on `null` in `getDeepMatches`

mnoorenberghe:
Community guidelines:

      and

This PR contains:

- [x] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [ ] no

Breaking Changes?

- [x] no

List any relevant issue numbers: #56 

### Description

`chrome.runtime.lastError` is `null` rather than `undefined`[1] in Firefox so the `getDeepMatches` caller in importWrapper--implicit.ts leads to an error being thrown.

This breaks the startup of my extension in Firefox.

[1] https://bugzilla.mozilla.org/show_bug.cgi?id=1406674

### DIFF
--- a/src/manifest-input/browser/getDeepMatches.test.ts
+++ b/src/manifest-input/browser/getDeepMatches.test.ts
@@ -1,0 +1,17 @@
+import { getDeepMatches } from './getDeepMatches'
+
+test("null keys don't cause an exception to be thrown when the predicate doesn't match", () => {
+  const obj = {
+    runtime: {
+      lastError: null,
+      onMessage: {
+        addListener() {},
+      },
+    },
+  }
+  expect(getDeepMatches(obj, (x) => typeof x === 'object' && 'addListener' in x, [])).toEqual([
+    {
+      addListener: obj.runtime.onMessage.addListener,
+    },
+  ])
+})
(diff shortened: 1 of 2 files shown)


mnoorenberghe:
The test failure doesn't seem related to my change. It's a TS issue related to the Cheerio module.

jacksteamdev:
@mnoorenberghe Looks good to me! Were you able to get Firefox working with this PR?

mnoorenberghe:
Yes, it gets much further but there are still other bugs to fix.

mnoorenberghe:
Without this fix none of the real background scripts would load IIRC, only a wrapper.

jacksteamdev:
What version of Firefox are you using?

This will be in the next version on NPM! I want to do some real world testing, so probably Tuesday.

mnoorenberghe:
I think it was Firefox Nightly version 90.0 but adding support for Firefox is a side-project so I haven't touched it since I opened this PR. 

The next issue I was looking into was that module imports in content scripts were getting resolved using the page's base URL rather than an extension base URL. I'm not sure whether Firefox or Chrome is correct in this regard (or perhaps it's undefined since extension content scripts aren't [standardized](https://browserext.github.io/browserext/) too much yet).

jacksteamdev:
We'll probably want to switch to creating script tags in the background page to get that side working.

The content script path should be fixable with `chrome.runtime.getURL`.

Looks like we're getting close to full Firefox support!

Thanks for the help, it's golden :medal_sports:

Issue #183, opened 2022-01-11
when i use vue of vite build extension.  that css will lose


yvywang:
### Build tool

Vite

### Where do you see the problem?

- [X] In the browser
- [ ] In the terminal

### Describe the bug

![image](https://user-images.githubusercontent.com/53296318/148920757-5f5b06dd-34e4-4e65-844b-efed6ca9aab2.png)

### Reproduction

![image](https://user-images.githubusercontent.com/53296318/148920921-8ebfb513-baff-49c5-9306-d05f9fa746f7.png)

### System Info

```shell
vite build
```

### Severity

blocking all usage of RPCE

yvywang:
I found that the CSS built was not being used

wudaxue:
I resolved it in the way shown in the figure。

![Wl4ZedvDT9](https://user-images.githubusercontent.com/31333358/149186373-e5ce7dd8-dce7-4387-9c67-67edb7d895bf.png)

by link   https://www.xiaoboy.com/topic/202109180735.html

jacksteamdev:
Are you saying that the CSS is not working? I need more info... :hugs: 

1. Please paste the result of running this command
```sh
npx envinfo --system --npmPackages rollup-plugin-chrome-extension,vite,rollup --binaries --browsers
```
2. Please paste the `manifest.json` from `outDir`

yvywang:
> ```shell
> npx envinfo --system --npmPackages rollup-plugin-chrome-extension,vite,rollup --binaries --browsers
> ```

  System:
    OS: macOS 12.0.1
    CPU: (10) arm64 Apple M1 Pro
    Memory: 241.83 MB / 16.00 GB
    Shell: 5.8 - /bin/zsh
  Binaries:
    Node: 16.13.1 - ~/.nvm/versions/node/v16.13.1/bin/node
    Yarn: 1.22.17 - /opt/homebrew/bin/yarn
    npm: 7.19.1 - ~/.nvm/versions/node/v16.13.1/bin/npm
  Browsers:
    Chrome: 97.0.4692.71
    Safari: 15.1
  npmPackages:
    rollup-plugin-chrome-extension: ^4.0.1-11 => 4.0.1-11
    vite: ^2.7.10 => 2.7.10
    --------------------------------------------------------------------------------------------------------
paste manifest.json from outDir 
    {
  "name": "weheatmap",
  "version": "1.0",
  "description": "Build an Extension!",
[...]

Issue #173, opened 2021-12-27
Uncaught TypeError: Cannot read properties of undefined (reading 'message')


i8ramin:
### Build tool

Vite

### Where do you see the problem?

- [X] In the browser
- [ ] In the terminal

### Describe the bug

Getting the following error when I run with the current beta version (with Vite support). Seems to be caused by this line of code here: https://github.com/extend-chrome/rollup-plugin-chrome-extension/blob/main/src/browser/code-runtimeReloader-csCode.ts#L19

```
runtime-reloader--content-script.js:1701 Uncaught TypeError: Cannot read properties of undefined (reading 'message')
    at Object.error (runtime-reloader--content-script.js:1701)
    at Object.error (runtime-reloader--content-script.js:462)
    at SafeSubscriber.Subscriber._error (runtime-reloader--content-script.js:408)
    at SafeSubscriber.Subscriber.error (runtime-reloader--content-script.js:386)
    at runtime-reloader--content-script.js:1381
    at OperatorSubscriber._this._error (runtime-reloader--content-script.js:626)
    at OperatorSubscriber.Subscriber.error (runtime-reloader--content-script.js:386)
    at OperatorSubscriber.Subscriber._error (runtime-reloader--content-script.js:408)
    at OperatorSubscriber.Subscriber.error (runtime-reloader--content-script.js:386)
    at runtime-reloader--content-script.js:963
```

Not entirely sure what the root cause is (could def be something in our app setup), but seems that `err` here is undefined. Any clues to try and debug this?

### Reproduction

This is happening within our local app and it is not something I can share a link for. 

### Logs

```shell
Uncaught TypeError: Cannot read properties of undefined (reading 'message')
    at Object.error (runtime-reloader--content-script.js:1701)
    at Object.error (runtime-reloader--content-script.js:462)
    at SafeSubscriber.Subscriber._error (runtime-reloader--content-script.js:408)
    at SafeSubscriber.Subscriber.error (runtime-reloader--content-script.js:386)
    at runtime-reloader--content-script.js:1381
    at OperatorSubscriber._this._error (runtime-reloader--content-script.js:626)
    at OperatorSubscriber.Subscriber.error (runtime-reloader--content-script.js:386)
    at OperatorSubscriber.Subscriber._error (runtime-reloader--content-script.js:408)
    at OperatorSubscriber.Subscriber.error (runtime-reloader--content-script.js:386)
    at runtime-reloader--content-script.js:963
````
[...]

PR #187, opened 2022-01-18
Special import types

jacksteamdev:
Community guidelines:

      and

This PR contains:

- [ ] bugfix
- [x] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [ ] no

Breaking Changes?

- [x] no

List any relevant issue numbers: possibly #183 

### Description

RPCE now supports the Chrome Scripting API, static asset imports in content scripts, and dynamic HTML files.

#### Script Imports and the Chrome Scripting API

Chrome recently added a new [Scripting API for MV3 Chrome Extensions](https://developer.chrome.com/docs/extensions/reference/scripting/#toc) that supports dynamic content scripts.

You can now import a script file with the `?script` query. RPCE will bundle the script, and the import will return a file path as the default export. This works in all script contexts: background, content, popup pages, etc.

```javascript
// background.ts
import scriptFilePath from "./dynamic-script.ts?script"
const tabId = getTabId();
chrome.scripting.executeScript(
    {
      target: {tabId: tabId},
      files: [scriptFilePath],
    },
    () => { ... });
```

If you need to inject a main-world script, import a script file into a content script and set it as the source for a script tag. This script will execute in the host page's main execution context, with access to global variables, etc. Make sure to get the Chrome Extension URL by calling `chrome.runtime.getURL`.

```javascript
// content-script.ts
import mainWorld from "./main-world-script.ts?script"
const src = chrome.runtime.getURL(mainWorld)
const script = document.createElement('script')
script.src = src
document.body.append(script)
```

The same web access restrictions apply to dynamic content scripts, so RPCE will add a new entry in `web_accessible_resources` when it finds a content script not declared in the manifest. The default match pattern is all HTTP and HTTPS pages. However, you can customize the match pattern by using the resource placeholder `<dynamic_scripts>` in `web_accessible_resources`:

```JSON
{
  "web_accessible_resources": [
    {
      "resources": ["<dynamic_scripts>"],
      "matches": ["https://google.com/*", "https://github.com/*"]
    }
  ]
}
```

Auto-reloading does not work on dynamic scripts yet, but that is coming soon.

#### HTML Imports

Sometimes it is necessary to include an HTML page in your extension, but there's nowhere to put it in the manifest, e.g., a devtools panel or an iframe in a content script. Now you can import an HTML page, and RPCE will bundle it.

```javascript
// content-script.ts
import html from "./iframe.html"
const src = chrome.runtime.getURL(html)
const iframe = document.createElement('iframe')
iframe.src = src
document.body.append(script)
```

#### Content Scripts and Web Accessible Resources

This PR adds content script support for any static asset import that Vite supports. If a content script uses an external image or CSS file, the manifest must declare those files as web accessible resources. RPCE analyzes the resources for each content script and adds those exact files to the manifest; only matching host pages can access those resources.


jacksteamdev:
This may fix #183 and #173

PR #272, opened 2022-04-04
Manifest resources and dynamic scripts

jacksteamdev:
Community guidelines:

      and

This PR contains:

- [x] bugfix
- [x] feature
- [x] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [ ] no

Breaking Changes?

- [x] no

### Description

Originally this PR was to add support for declaring scripts in `web_accessible_resources`. While a logical place to declare scripts, this approach proved fairly complex and led to more config options than I wanted. 

We already have a Dynamic Script API that serves this use case, and this PR fixes some regressions with that API.

### Usage

The Chrome Scripting API supports dynamic content scripts, which can be executed or registered from the background.

RPCE supports the Scripting API with the `?script` query, which bundles the imported script as a content script and returns the output script path. During development, an imported dynamic script will support HMR in the same way as a content script declared in the manifest.

```js
// background.ts
import scriptPath from './dynamic-script?script'

const [tabId] = await chrome.tabs.query({ active: true, currentWindow: true })
const result = await chrome.scripting.executeScript({
  target: { tabId },
  files: [scriptPath],
})
```

Scripts using the `?script` query use a dynamic import wrapper script, which is required to support ES modules in content scripts. The imported script path is actually the path to the wrapper script. This means it is only appropriate for dynamic content scripts.

If you need to use a content script to inject a main world script into the host page's main execution context, you can use the `?script&module` query to import the path of a plain ES module output file. Main world scripts are helpful to decorate object methods that the host page will use.

```js
// content-script.ts
import scriptPath from './main-world?script&module'

const script = document.createElement('script')
script.src = chrome.runtime.getURL(scriptPath)
script.type = "module"
```

### Why not `web_accessible_resources`?

Supporting `web_accessible_resources` introduces several edge cases. What if a resource uses a glob pattern that matches `node_modules`? What file name will an output script from `web_accessible_resources` use? What if a content script should be a different format (IIFE, ESM, etc)? I want to avoid extending the already dense manifest format, and I also want to avoid adding unnecessary config options for RPCE.

### Future of Imported Script API

The Dynamic Script API will eventually cover three use cases: ESM content scripts, IIFE content scripts, and main world content scripts. All three are essentially the same in usage, but they cover different scenarios and use different output formats:

- ESM content scripts use an import wrapper to support code-splitting, HMR, and all their imported resources go into `web_accessible_resources`.
- Main world scripts themselves are `web_accessible_resources` of the importing content script, and don't use an import wrapper.
- IIFE content scripts: `web_accessible_resources` don't work everywhere, so we need a non-code split content script option. (see Chromium bug [1219825 - Manifest V3: Opaque origins can't embed web accessible extension resources](https://bugs.chromium.org/p/chromium/issues/detail?id=1219825&q=component%3APlatform%3EExtensions%20opaque%20origin&can=2))

Related to #269


emred2700:
Hi, thanks for this. I was pulling my hair on this until I remembered checking the issues. This should be in the documentation :) Great job @jacksteamdev

Issue #279, opened 2022-04-13
build fails when CSP contains the 'self' directive


jpozo20:
### Build tool

Vite

### Where do you see the problem?

- [ ] In the browser
- [X] In the terminal

### Describe the bug

When the manifest.json contains the 'self' directive either in content_security_policy.extension_pages or content_security_policy.sandbox, then the `vite build` fails with `Error: unable to parse manifest code`. 

### Reproduction

https://github.com/jpozo20/tailwind-bug/tree/csp-bug

### Logs

```shell
➜ yarn build --debug
yarn run v1.22.15
$ vite build --debug
  vite:config bundled config file loaded in 343.70ms +0ms
  vite:config using resolved config: {
  vite:config   plugins: [
  vite:config     'vite:build-metadata',
  vite:config     'alias',
  vite:config     'vite:react-babel',
  vite:config     'vite:react-refresh',
  vite:config     'crx:content-scripts-pre',
  vite:config     'crx:dynamic-scripts-load',
  vite:config     'crx:manifest-init',
  vite:config     'crx:manifest-loader',
  vite:config     'crx:stub-input',
  vite:config     'vite:modulepreload-polyfill',
  vite:config     'vite:resolve',
  vite:config     'vite:html-inline-proxy',
  vite:config     'vite:css',
  vite:config     'vite:esbuild',
  vite:config     'vite:json',
  vite:config     'vite:wasm',
  vite:config     'vite:worker',
  vite:config     'vite:asset',
  vite:config     'crx:dynamic-scripts-build',
  vite:config     'vite:define',
  vite:config     'vite:css-post',
  vite:config     'vite:build-html',
  vite:config     'vite:worker-import-meta-url',
  vite:config     'vite:watch-package-data',
  vite:config     'commonjs',
  vite:config     'vite:data-uri',
  vite:config     'rollup-plugin-dynamic-import-variables',
  vite:config     'vite:asset-import-meta-url',
  vite:config     'crx:hmr',
  vite:config     'crx:content-script-resources',
  vite:config     'crx:content-scripts-post',
  vite:config     'crx:background-loader-file',
  vite:config     'vite:build-import-analysis',
  vite:config     'vite:esbuild-transpile',
  vite:config     'vite:terser',
  vite:config     'vite:manifest',
  vite:config     'vite:reporter',
  vite:config     'vite:load-fallback',
  vite:config     'crx:manifest-post'
  vite:config   ],
[...]

PR #281, opened 2022-04-13
fix: vite build issue when using self directive in manifest.json

jpozo20:
Community guidelines:

      and

This PR contains:

- [x] bugfix
- [ ] feature
- [ ] refactor
- [ ] documentation
- [ ] other

Are tests included?

- [ ] no

Breaking Changes?

- [x] no

List any relevant issue numbers: #279 

### Description

When the manifest.json contains the 'self' directive either in content_security_policy.extension_pages or content_security_policy.sandbox, then `vite build` fails with `Error: unable to parse manifest code`. 

After debugging for a while, I found out this is because the manifest is transformed into a file with a named default export I guess because of the single quotes needed in the CSP. This was causing the node tree to not contain a `Literal` as the `decodeManifest` function expects, instead the value of the json was found inside a `TemplateElement`.  

The PR adds an `AcornTemplateElement` that is verified if the value from the Literal comes empty and throws the exception if the template element is also empty. It was my first time fixing something in a node a library so it took me a while to setup the environment to actually debug. I will add tests covering the error and the solution later.

### DIFF
--- a/tests/mv3/vite-self-directive-in-csp/build.test.ts
+++ b/tests/mv3/vite-self-directive-in-csp/build.test.ts
@@ -1,0 +1,17 @@
+import { build } from 'tests/runners'
+import { testOutput } from 'tests/testOutput'
+import { mockDate } from 'tests/helpers'
+
+mockDate()
+
+describe('Test manifest v3', () => {
+  it("should work if the 'self' directive is not found", async () => {
+    const result = await build(__dirname)
+    await testOutput(result)
+  })
+
+  it("should fail if the 'self' directive is found", async () => {
+    const runBuild = async () => await build(__dirname, 'vite.csp.ts')
+    await expect(runBuild).rejects.toThrowError('unable to parse manifest code')
+  })
+})
--- a/src/node/helpers.ts
+++ b/src/node/helpers.ts
@@ -13,6 +13,7 @@ import type {
   AcornIdentifier,
   AcornLiteral,
   AcornMemberExpression,
+  AcornTemplateElement,
   ManifestFiles,
 } from './types'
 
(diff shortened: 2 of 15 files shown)


jacksteamdev:
@jpozo20 Thanks for digging in and contributing 🤗 

LGTM! A very simple snapshot test in `tests/mv3` should be enough. If it passes locally, don't worry about CI passing. Some of the serve tests are flaky.

jpozo20:
@jacksteamdev I copied the basic-ts folder and added two tests. I don't really get how snapshots work outside of components, but the second test will fail with the new code since it expects to throw an error if the 'self' directive is found, while both will work if you comment the assignment of the `templateElement` in the `simple` visitor since RPCE will throw the error.

jacksteamdev:
@jpozo20 Yeah, it's a kind of alternative use of snapshot testing. We need to add a section on tests in the upcoming docs. Basically, the snapshot testing strategy validates the output files and fails when the output of RPCE changes. 

The functions `build` and `serve` run the [Vite JavaScript API](https://vitejs.dev/guide/api-javascript.html) and perform some setup and cleanup tasks. They will fail if Vite throws an error.

The function `testOutput` checks that all the files in the manifest are present in the output files, and uses snapshots to verify that the list of output files has not changed and that the contents of the output files did not change.

I treat the snapshot files as code. If a change shows up in Git that I don't expect, I know that my changes are causing some unintended side effects. Previously I was using a more traditional testing strategy with explicit expectations, but that failed in two ways: unexpected changes would slip through my tests, and it took a loooong time to update tests to handle changes. With snapshots, it's easy to know when something changes, and easy to update tests when those changes are correct.

tests/mv3/vite-self-directive-in-csp/build.test.ts
```diff
@@ -0,0 +1,17 @@
...
+  it("should work if the 'self' directive is not found", async () => {
+    const result = await build(__dirname)
+    await testOutput(result)
+  })
+
+  it("should fail if the 'self' directive is found", async () => {
+    const runBuild = async () => await build(__dirname, 'vite.csp.ts')
+    await expect(runBuild).rejects.toThrowError('unable to parse manifest code')
+  })
+})
```
jacksteamdev, line 17: We just need one test that passes if the 'self' directive is present. The original `basic-ts` test handles the case where there is no CSP.

jacksteamdev (changes requested):
Overall this looks good! Hopefully, I was able to explain the testing strategy successfully.

jacksteamdev:
I'm getting ready to make some big changes to the structure of this repo, so I'm going to merge this PR. Thanks for chasing down this bug and making the fix!

Issue #293, opened 2022-04-14
MV3: Invalid path separator in the generated manifest.json on Windows


nerblock:
### Build tool

Rollup

### Where do you see the problem?

- [X] In the browser
- [ ] In the terminal

### Describe the bug

When building a project for MV3 with https://github.com/extend-chrome/ts-react-boilerplate/ on a Windows machine the content script can't be loaded, because the path separators in the `web_accessible_resources` section in the generated `dist/manifest.json` file are incorrecty converted to Windows style path separators:

```
  "web_accessible_resources": [
    {
      "resources": [
        "chunks\\*-*.js",
        "content\\index.js"
      ],
      "matches": [
        "https://*/*",
        "http://*/*"
      ]
    }
  ]
```

### Expected behavior

manifest.json should contain Unix style path separators:

```
  "web_accessible_resources": [
    {
      "resources": [
        "chunks/*-*.js",
        "content/index.js"
      ],
      "matches": [
        "https://*/*",
        "http://*/*"
      ]
    }
  ]
```

### Actual behavior

Running `npm run build` on a Linux machine (or in WSL) generates `dist/manifest.json` with `/` (slash) as path separator.

Running `npm run build` on a Windows machine generates `dist/manifest.json` with `\` (backslash) as path separator.

### Reproduction

1. clone the repository https://github.com/extend-chrome/ts-react-boilerplate/ on a Windows machine
2. Modify `src/manifest.json` for MV3:
```
{
  "name": "mv3",
  "manifest_version": 3,
  "content_scripts": [
    {
      "js": [
        "content/index.ts"
      ],
      "matches": [
        "https://*/*",
        "http://*/*"
      ]
    }
  ]
  "action": {
    "default_title": "Popup",
    "default_popup": "pages/popup/index.html"
  }
}
```
4. open a powershell terminal
5. run `npm run build`
6. check the contents of dist/manifest.json

### System Info

```shell
  System:
    OS: Windows 10 10.0.19043
    CPU: (12) x64 Intel(R) Core(TM) i7-5820K CPU @ 3.30GHz
    Memory: 1.30 GB / 15.91 GB
  Binaries:
    Node: 16.14.0 - C:\Program Files (x86)\nodejs\node.EXE
    npm: 8.3.1 - C:\Program Files (x86)\nodejs\npm.CMD
  Browsers:
    Chrome: 100.0.4896.88
[...]

PR #308, opened 2022-04-14
Use slash to guarantee support Windows

nerblock:
Community guidelines:

      and

This PR contains:

- [x] bugfix

Are tests included?

- [ ] yes (_bugfixes and features will not be merged without tests_)
- [x] no, because in `npm run test` every test fails with this error: `TypeError: Jest: a transform must export a 'process' function.`. `npm install` doesn't work either, I needed to run `npm install --legacy-peer-deps` to successfully install the dependencies. I don't want to fix the jest config and package.json just for this little patch. Please run the tests yourself.

Eslint completed without error.

I did not include the output of prettier, because it changed other parts of the code.

Also I replicated your comment from the MV2 code path to describe what this fix does. (Use slash to guarantee support Windows)

Breaking Changes?

- [x] no

#293

### Description

This change guarantees that only posix style path separators are present in the web_accessible_resources section of the generated manifest.json file on Windows.

### DIFF
--- a/src/manifest-input/updateManifest.ts
+++ b/src/manifest-input/updateManifest.ts
@@ -1,4 +1,5 @@
 import { basename, relative } from 'path'
+import slash from 'slash'
 import { RollupOptions } from 'rollup'
 import { ManifestInputPluginCache } from '../plugin-options'
 import { cloneObject } from './cloneObject'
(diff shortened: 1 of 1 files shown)


jacksteamdev:
Thanks for the PR! I'm currently in the process of converting this repo to a monorepo that will have both RPCE@3.6.7 and RPCE@beta as two packages. I'll ping you when it's ready and you can update this PR to match the new project layout.

Issue #288, opened 2022-04-13
HMR doesn't replace tailwind css


riywo:
### Build tool

Vite

### Where do you see the problem?

- [X] In the browser
- [ ] In the terminal

### Describe the bug

Related to #274 , but I'm not sure it's the same issue. I'll spend more time to reproduce later and update this description. Sorry for a poor report now.

When I update a tailwind class in a React component, RPCE reloads components and its class name is updated. However, the css embedded by `content_scripts` isn't updated until I restart `vite`. Therefore, the new class isn't accessible even I reload the target page by browser because tailwind postcss? only imports the classes actually used.

By checking Chrome devtools, it looks like triggering some update on the embedded css but the content is still old.

When I access the `vite` server directly, though, the css embedded is updated correctly (the new class is appended). So, I guess the signal is sent from `vite` server but RPCE misses it somewhere. 

I can workaround by using `nodemon` to restart the `vite` server whenever files are changed, but it's slower than HMR.

### Reproduction

TODO

### Logs

```shell
Will update once created a repro.
```

### System Info

```shell
Will update once created a repro.
```

### Severity

annoyance

liaoliao666:
I have also this issue, its a pain point to me. hope it gets fixed soon.

jacksteamdev:
This should be fixed in v1.0.6 :rocket: [@crxjs/vite-plugin - npm](https://www.npmjs.com/package/@crxjs/vite-plugin)

Issue #274, opened 2022-04-06
Issues with hmr when using tailwind.css


jpozo20:
### Build tool

Vite

### Where do you see the problem?

- [X] In the browser
- [X] In the terminal

### Describe the bug

There are some issues when using tailwind.css along RPCE.

1. When using tailwind.css during dev, during hot reload it will try to reload index.css multiple times as well as any file that imports the index.css, causing a max stack exceeded error. This can be solved by changing the `isImporter` function in the `src/node/plugin-hmr.ts` to the code below, but issues 2 and 3 still persist.

```javascript
function isImporter(file) {
	const seen = [];
	const pred = (node) => {
		if (node.file === file) {
			return true;
		}
		seen.push(node);
		for (const node2 of node.importers) {
			if (seen.indexOf(node) < 0) {
				if (pred(node2)) {
					return true;
				}
			}
		}
	};
	return pred;
}
```

2. If you only import the index.css in the entry point of the project, then it will always try to load that CSS file when you save any other file. 
![reloading-index-css](https://user-images.githubusercontent.com/9166182/162005500-f7701d14-7660-4668-ba1e-6b071c773b2f.png)

3. If you import a .module.css file, then it gets totally crazy trying to reload the index.css and unrelated (not modified) files. Though in this case Spinner.tsx is a child of app.tsx, so I don't know if it's related
![crazy-reload](https://user-images.githubusercontent.com/9166182/162006090-4b19fdf3-5d97-4e75-917e-c95e76db16fc.png)

### Reproduction

[https://github.com/jpozo20/tailwind-bug](https://github.com/jpozo20/tailwind-bug)

### Logs

```shell
<details>
<summary>yarn start --debug</summary>
➜ yarn start --debug
yarn run v1.22.15
$ vite --debug
  vite:config bundled config file loaded in 450.68ms +0ms
  vite:config using resolved config: {
  vite:config   plugins: [
  vite:config     'crx:html-auditor-pre',
  vite:config     'vite:pre-alias',
  vite:config     'alias',
  vite:config     'vite:react-babel',
  vite:config     'vite:react-refresh',
  vite:config     'crx:hmr',
  vite:config     'crx:dynamic-scripts',
  vite:config     'crx:manifest-init',
  vite:config     'vite:modulepreload-polyfill',
  vite:config     'vite:resolve',
  vite:config     'vite:optimized-deps',
  vite:config     'vite:html-inline-proxy',
  vite:config     'vite:css',
[...]

PR #385, opened 2022-05-25
Fix HMR bugs

jacksteamdev:
This PR includes a couple of quality-of-life fixes for HMR.

1. Automatically ignores `build.outDir` for server, so HMR doesn't trigger a full reload when the file writer writes.
2. Fixes flaky HMR updates for content scripts; Tailwind should work fine now.


jacksteamdev:
Closes #274

jacksteamdev:
Closes #288

101arrowz:
Hey @jacksteamdev, sorry to barge in on this PR but I'm working on Parcel's implementation of Manifest V3 HMR and I'm wondering how you're getting around the Chromium bug causing `localhost` to be banned in content security policy.

I noticed you added [this section](https://github.com/crxjs/chrome-extension-tools/blob/a1e27280c5ec8e2b2ad63a4995f8d3bcdf9758d4/packages/vite-plugin/src/client/es/hmr-client-worker.ts#L18-L51) to force the content scripts, options pages, etc. to fetch from the server rather than from the extension on the disk, but when you hot-reload the background service worker the fetch interceptor obviously doesn't work. In fact as far as I can tell the only way to download dynamic scripts in background SW is to `importScripts(url)` (or `import(url)` for module service workers, though this is also broken because `import()` is for now still banned in service workers). Neither of these can be intercepted.

So is it true that for the time being background service worker hot reload is impossible?

jacksteamdev:
> So is it true that for the time being background service worker hot reload is impossible?

@101arrowz As I understand it, module imports for service workers are different; it's kind of synchronous, so async things like dynamic import and top-level await don't work. I'm using `chrome.runtime.reload()` to force a full extension reload for background changes.

PR #421, opened 2022-06-11
Fix link to Extend Chrome Docs in README.md

tadwohlrapp:
Current link pointed to extend-chrome.com instead of extend-chrome.dev


jacksteamdev:
@tadwohlrapp Thanks!

Issue #389, opened 2022-05-30
`default_icon` in `manifest.action` does not support string


Chersquwn:
### Build tool

Vite

### Where do you see the problem?

- [ ] In the browser
- [X] In the terminal

### Describe the bug

It seems `default_icon` in `action` does not support string. If `default_icon` uses string, it will be split into `'i','c','o','n','s','/'...`, such as `icons/`.

https://github.com/crxjs/chrome-extension-tools/blob/main/packages/vite-plugin/src/node/helpers.ts#L71
<img width="558" alt="image" src="https://user-images.githubusercontent.com/8766318/170948529-aaa6eb69-2563-44dc-bca7-e7bf1134e180.png">

### Reproduction

```js
// manifest.json
{
  "name": "Demo",
  "version": "1.0.0",
  "manifest_version": 3,
  "action": {
    "default_title": "Demo Title",
    "default_icon": "icons/logo-128.png"
  },
  "icons": {
    "16": "icons/logo-16.png",
    "48": "icons/logo-48.png",
    "128": "icons/logo-128.png"
  }
}
```

### Logs

```shell
[crx:manifest-post] ENOENT: Could not load manifest asset "i".
Manifest assets must exist in one of these directories:
```

### System Info

```shell
System:
    OS: macOS 12.3.1
    CPU: (16) x64 Intel(R) Core(TM) i9-9880H CPU @ 2.30GHz
    Memory: 80.28 MB / 16.00 GB
    Shell: 5.8 - /bin/zsh
  Binaries:
    Node: 14.19.0 - ~/.nvm/versions/node/v14.19.0/bin/node
    Yarn: 1.22.18 - /usr/local/bin/yarn
    npm: 6.14.16 - ~/.nvm/versions/node/v14.19.0/bin/npm
  Browsers:
    Chrome: 102.0.5005.61
    Safari: 15.4
  npmPackages:
    rollup-plugin-chrome-extension: 4.0.1-25 => 4.0.1-25 
    vite: ^2.9.9 => 2.9.9
```

### Severity

annoyance

filipw01:
@Chersquwn string for default icon is not supported in manifest v3, see https://json.schemastore.org/chrome-manifest. @jacksteamdev I guess the manifest validation could pick it up though right?

Chersquwn:
@filipw01 But I don't  see any migration guide about "default icon not support string type" on the chrome developer website.In the official examples, there are also default icons that use string.
Like this,
[...]

PR #424, opened 2022-06-12
Fix manifest icons type

holazz:
- icons can be a string


jacksteamdev:
@holazz Thanks, LGTM!

Closes #389

PR #463, opened 2022-08-11
Fix defineManifest example in docs

anatolykopyl:
Example from the docs doesn't work, since `defineManifest` only accepts one parameter:

```typescript
// defineManifest.ts
export type ManifestV3Fn = (env: ConfigEnv) => ManifestV3 | Promise<ManifestV3>
```


jacksteamdev:
@anatolykopyl Thanks!

PR #475, opened 2022-08-19
docs: fix manifest ts example regex

ryoid:
Example from the docs has an issue with the regex not replacing all the non digits, dots, or dash characters.


jacksteamdev:
@ryoid Great catch! 🥇 Thanks

Issue #455, opened 2022-07-28
The paths specified in the manifest are not resolved correctly


GoodSimonVM:
### Build tool

Vite

### Where do you see the problem?

- [ ] In the browser
- [X] In the terminal

### Describe the bug

I have configured `rootDir` in `tsconfig.json` and `root` in `vite.config.ts`, but in the manifest I still have to write paths relative to the root of the project. If I specify a path relative to the root specified in the configuration, for example `"default_popup": "pages/popup/index.html"` (without `src/`), then I get an error:
```sh
Could not resolve entry module (pages/popup/index.html).
error during build:
Error: Could not resolve entry module (pages/popup/index.html).
    at error (file:///D:/Projects/chrome-extension/node_modules/rollup/dist/es/shared/rollup.js:1858:30)
    at ModuleLoader.loadEntryModule (file:///D:/Projects/chrome-extension/node_modules/rollup/dist/es/shared/rollup.js:22352:20)
    at async Promise.all (index 0)
```
If I specify the path along with `src/`, then after the build my `src/pages/popup/index.html` will be placed in `dist/pages/popup/index.html`, but the path in the output `manifest.json` file will be `src/pages/popup/index.html`. So this is a error, because the path `dist/src/pages/popup/index.html` doesn't exist and `chrome` can't load this popup page.

### Reproduction

Use this minimal example: [chrome-extension.zip](https://github.com/crxjs/chrome-extension-tools/files/9209976/chrome-extension.zip)

### Logs

```sh
PS D:\Projects\chrome-extension> npm run build
npm WARN config global `--global`, `--local` are deprecated. Use `--location=global` instead.

> chrome-extension@0.0.1 build
> vue-tsc --noEmit && vite build

vite v3.0.3 building for production...
✓ 2 modules transformed.
Could not resolve entry module (pages/popup/index.html).
error during build:
Error: Could not resolve entry module (pages/popup/index.html).
    at error (file:///D:/Projects/chrome-extension/node_modules/rollup/dist/es/shared/rollup.js:1858:30)
    at ModuleLoader.loadEntryModule (file:///D:/Projects/chrome-extension/node_modules/rollup/dist/es/shared/rollup.js:22352:20)
    at async Promise.all (index 0)
```

### System Info

```shell
System:
  OS: Windows 10 10.0.19042
  CPU: (12) x64 Intel(R) Core(TM) i5-10500 CPU @ 3.10GHz
[...]

PR #427, opened 2022-06-14
File writer refactor

jacksteamdev:
It's time to wind up the creation phase of CRXJS, and the file writer needs a better design. I have an idea for an HMR-based file writer that excludes the need for Rollup, or any kind of watcher.

Currently, the file writer is the most obtuse part of the project, with a confusing event-based model. Because it produces a full build for every change, even with caching it uses more resources than necessary.

This rewrite won't change any exposed API details, and with the existing tests, it shouldn't cause any significant regressions.

## Specs

1. Track output scripts with `file-writer-scripts`
	1. `type ScriptModule = Promise<{type, viteUrl, filename}>` 
	2. `type ScriptType = 'loader'|'module'|'iife'|'dep'`
	3. `start(server: ViteDevServer): Promise<void>`
	4. `add(viteUrl: string, options?: {type: ScriptType}): ScriptModule`
	5. `update(viteUrl: string, options: {type: ScriptType}): Promise<void>`
	6. `reset(): Promise<void>` Removes all content scripts and deletes files.
	7. `modules` Module graph as `Map<ScriptModule>`
		1. by Vite URL
	8. `viteUrlToFilename(viteUrl: string): string` Convert Vite URL to output filename
		1. `/\?/g -> '__'`
		2. `/\&/g -> '--'`
		3. replacedUrl + `.js`
2. Call `start(server)` to start file writer
	1. File writer will output scripts as Rollup and Vite call `add()`
	2. No need to wait until pre-bundling is complete
	3. Need to queue scripts in case pre-bundling starts before file writer
	4. Need to restart file writer when server restarts
3. Run Rollup to output the bones of the extension
	1. Emit manifest
	2. Emit html placeholders
	3. Emit script loaders
	4. Emit manifest assets
	5. Emit public assets
4. `add` content scripts from manifest
5. `add` content scripts from `?script` imports
6. On call to  `add` (sometimes no-op)
	1. Check for file in module graph
	2. If file in graph, resolve module
	3. Else, `update` file
7. On call to `update` (always rewrite file)
	1. Transform file, then concurrently
		1. Output to file system
		2. Recursively `add` file deps
	2. Resolve void 
8. Output types
	1. `script&loader` Use content script loader to gain HMR
	2. `script&module` Output raw ES module
		1. Maybe don't need this; does Vite support `?module`?
	3. `script&iife` Output IIFE format
		1. Use Worker API

## Notes

[...]

PR #517, opened 2022-09-28
fix: fixed problem paths not resolving correctly

lucasfernandodev:
Problem #455 was bothering me a lot in the project I was developing, so I decided to try to fix it. I'm new to programming and this is my first contribution to a project.


ralphschindler:
Hi, I am having the same problem with custom `root`, does this fix that problem and can it be included in vite-plugin when using vite2?

lucasfernandodev:
Yes, but if I'm not mistaken in one of the last commits was to add a new File Write (#427) that should solve this problem.

Issue #511, opened 2022-09-07
[Bug] build error when specifying '<all_urls>' in 'matches' field of 'content_scripts' in the manifest


bruceeewong:
### Build tool

Vite

### Where do you see the problem?

- [ ] In the browser
- [X] In the terminal

### Describe the bug

when setting <all_urls> in matches field of content script in the manifest json
```
// manifest.json
{
  content_scripts: [
    {
      matches: ['<all_urls>'],
      js: ['src/scripts/content/index.ts'],
    },
  ],
}
```

it triggers errors during building:

```
[crx:content-script-resources] TypeError: Cannot read properties of undefined (reading 'split')

// error stack at @crxjs/vite-plugin/dist/index.mjs:104:35)

const stubMatchPattern = (pattern) => {
  const [schema, rest] = pattern.split("://");  // ! <- the split logic is the problem
  const [origin, pathname] = rest.split("/");
  const root = `${schema}://${origin}`;
  return pathname ? `${root}/*` : root;
};
```

### Reproduction

set manifest like bellow, and run vite build
```
// manifest.json
{
  content_scripts: [
    {
      matches: ['<all_urls>'],  // <- set all_urls would trigger the building error
      js: ['src/scripts/content/index.ts'],
    },
  ],
}
```

### Logs

```shell
[crx:content-script-resources] TypeError: Cannot read properties of undefined (reading 'split')
    at stubMatchPattern (file:///Users/bruski/workspace/projects/suiet/suiet/node_modules/@crxjs/vite-plugin/dist/index.mjs:104:35)
    at Array.map (<anonymous>)
    at Object.renderCrxManifest (file:///Users/bruski/workspace/projects/suiet/suiet/node_modules/@crxjs/vite-plugin/dist/index.mjs:3322:59)
    at Object.generateBundle (file:///Users/bruski/workspace/projects/suiet/suiet/node_modules/@crxjs/vite-plugin/dist/index.mjs:2919:60)
    at async Bundle.generate (file:///Users/bruski/workspace/projects/suiet/suiet/node_modules/rollup/dist/es/shared/rollup.js:15990:9)
    at async file:///Users/bruski/workspace/projects/suiet/suiet/node_modules/rollup/dist/es/shared/rollup.js:23754:27
    at async catchUnfinishedHookActions (file:///Users/bruski/workspace/projects/suiet/suiet/node_modules/rollup/dist/es/shared/rollup.js:23088:20)
[...]

PR #661, opened 2023-03-09
fix all_urls

tomdygo:
Currently the all_urls match pattern causes builds to fail. 

This is just a fix that was introduced for the vite plugin:
https://github.com/crxjs/chrome-extension-tools/issues/511
https://github.com/crxjs/chrome-extension-tools/pull/460

### DIFF
--- a/.changeset/late-paws-change.md
+++ b/.changeset/late-paws-change.md
@@ -1,0 +1,5 @@
+---
+'rollup-plugin-chrome-extension': patch
+---
+
+Allows <all_urls> match in manifest.json within rollup package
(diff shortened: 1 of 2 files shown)


tomdygo:
@jacksteamdev - Sorry for the nudge in advance, this is blocking a build for us at the moment so hoping you'll be able to review :)

jacksteamdev:
@tomdygo Thanks!

tomdygo:
@jacksteamdev  - I think something must have caused the npm build to fail because the package appears to be empty now?
image

https://www.npmjs.com/package/rollup-plugin-chrome-extension?activeTab=explore

PR #670, opened 2023-03-15
fix: schema of input_components

t123yh:
The current definition of input_components is outdated. Update them to match [the official document](https://developer.chrome.com/docs/extensions/mv3/manifest/input_components/). Only v3 is updated because I cannot find v2 document anymore.


jacksteamdev:
@t123yh Thanks for the PR!

PR #701, opened 2023-05-04
Add inline sourcemap support to content scripts

jacksteamdev:
This PR adds inline sourcemap support to content scripts.


andresgutgon:
Is this already published. I see this in development
image


These are my versions:
`@crxjs/vite-plugin: 2.0.0-beta.17`
`@vitejs/plugin-react: ^4.0.0`
`vite: 4.3.9`

PR #711, opened 2023-05-08
add offscreen to llist of permissions

eduardhasanaj:
I added offscreen permission but unfortunately could not test it.

### DIFF
--- a/schema/manifest-v3.schema.json
+++ b/schema/manifest-v3.schema.json
@@ -166,7 +166,8 @@
           "wallpaper",
           "webNavigation",
           "webRequest",
-          "webRequestBlocking"
+          "webRequestBlocking",
+          "offscreen"
         ],
         "type": "string"
       },


filipw01:
I think it would be great to add all the other missing options in one go. From a quick look I noticed that `declarativeNetRequestWithHostAccess` is missing and there are probably more https://developer.chrome.com/docs/extensions/mv3/declare_permissions/

eduardhasanaj:
@filipw01 if you know how to build the packages in this monorepo please tell me.
Actually for offscreen, we need to modify bundling process so ofscreen html and ts are included in the dist.

filipw01:
Sadly I don't know and don't really have much time to dig into it. I didn't know modification of the bundling process is necessary?

I would suspect this should work without any changes to the plugin, are there issues with that?

```ts
import html from "./off_screen.html"

chrome.offscreen.createDocument({
  url: chrome.runtime.getURL(html),
  reasons: ['CLIPBOARD'],
  justification: 'reason for needing the document',
});

chrome.offscreen.closeDocument()
```

eduardhasanaj:
@filipw01 thank you very much for your efforts. 
You gave me  a new direction.
It almost worked.
Inside offscreen html I have a script tag which is not being compiled and throws following error:
```
RollupError: Unexpected token (Note that you need plugins to import files that are not JavaScript)
```

My offscreen html:
```
<script type="module" src="./offscreen.ts"></script>
```
The exact same line is used in popup html and here the script is compiled.
I am using rollup.

eduardhasanaj:
I am closing this as my changes are not added to rollup plugin.

Issue #627, opened 2023-01-15
Extra HTML pages don't find the src file in dev


corrortiz:
### Build tool

Vite

### Where do you see the problem?

- [X] In the browser
- [X] In the terminal

### Describe the bug

When you run the dev server the the extra HTML don't generate the reference to the source file, when you run the build command you get the correct reference:

### Output with the *vite build* command (this works correctly):
```
<head>
  <meta charset="UTF-8" />
  <meta name="viewport" content="width=device-width, initial-scale=1.0" />
  <meta http-equiv="X-UA-Compatible" content="ie=edge" />
  <title>OffScreen</title>
  <script type="module" crossorigin src="/assets/offscreen.7c2aa4e4.js"></script>
  <link rel="modulepreload" crossorigin href="/assets/chunk-1ebdbb1b.js">
  <link rel="modulepreload" crossorigin href="/assets/chunk-641ab1c8.js">
  <link rel="modulepreload" crossorigin href="/assets/chunk-2a3b33e1.js">
  <link rel="modulepreload" crossorigin href="/assets/chunk-9744cff8.js">
  <link rel="modulepreload" crossorigin href="/assets/chunk-745e3ca4.js">
</head>
<body>
  <div id="root"></div>
</body>
</html>

```
### Output with the *vite build* command:
```
<head>
  <meta charset="UTF-8" />
  <meta name="viewport" content="width=device-width, initial-scale=1.0" />
  <meta http-equiv="X-UA-Compatible" content="ie=edge" />
  <title>OffScreen</title>
</head>

<body>
  <div id="root"></div>
  <script src="./index.ts" type="module"></script>
</body>
</html>
``` 
I try to add the file with `rollup-plugin-copy` but its not working as it does not generate the missing file 

### Reproduction

N/A

### System Info

```shell
System:
    OS: macOS 13.1
    CPU: (8) x64 Intel(R) Core(TM) i7-8569U CPU @ 2.80GHz
    Memory: 36.57 MB / 16.00 GB
    Shell: 5.9 - /usr/local/bin/zsh
  Binaries:
    Node: 16.16.0 - ~/.nvm/versions/node/v16.16.0/bin/node
    Yarn: 1.22.19 - /usr/local/bin/yarn
    npm: 8.19.2 - ~/.nvm/versions/node/v16.16.0/bin/npm
  Browsers:
    Chrome: 109.0.5414.87
    Firefox: 108.0.2
    Safari: 16.2
```

### Severity

[...]

PR #713, opened 2023-05-10
Added support for offscreen document

eduardhasanaj:


jacksteamdev:
@eduardhasanaj Thanks for the PR! We certainly need to add `offscreen` to the permissions. 

Two considerations:
- I'm not sure that we need a new plugin option. Did you try the CRXJS instructions for [extra HTML pages](https://crxjs.dev/vite-plugin/concepts/pages#extra-html-pages)? If so, does adding it through the options fix it?
- Offscreen docs are a bit tricky, IIRC Chrome renders them weirdly in vite dev mode. I'd _reeeallly_ like some e2e tests for this feature before merging this PR.

GWSzeto:
> > I'm not sure that we need a new plugin option. Did you try the CRXJS instructions for [extra HTML pages](https://crxjs.dev/vite-plugin/concepts/pages#extra-html-pages)? If so, does adding it through the options fix it?

This method works when doing a build into preview, but the `offscreen.html` doesn't seem to be included in the `dist` when using `npm run dev`

The docs around [extra HTML pages](https://crxjs.dev/vite-plugin/concepts/pages) indicate that it would include them during development, but that doesn't seem to be the case

DblK:
For me, I wanted to play a sound using offscreen page. Without interaction inside a standard iframe, sound won't be played.
Using `offscreen` should do the trick.
Hope to see it soon!

GWSzeto:
@DblK Hope you're able to find a solution for this soon. In case you're interested, this is what worked to get it to work in dev

https://github.com/crxjs/chrome-extension-tools/issues/627#issuecomment-2019493008

sanzenwin:
@jacksteamdev Is there any news? We really need this feature.

Dozorengel:
@jacksteamdev looking forward for any news, I tried all the workarounds, nothing works properly

Issue #719, opened 2023-05-26
Infinite recursion on build due to circular references in JS chunks


fractalo:
### Build tool

Vite

### Where do you see the problem?

- [ ] In the browser
- [X] In the terminal

### Describe the bug

If the js chunk files created by building the content script by vite have circular references,
Unable to build with a Maximum call stack size exceeded error due to infinite recursion of the compileFileResources function.

The issue can be fixed by modifying the compileFileResources function to not reprocess files that have already been processed. I'll submit a PR for this.

### Reproduction

https://github.com/fractalo/crxjs-vite-dynamic-import

### Logs

```shell
[crx:web-accessible-resources] RangeError: Maximum call stack size exceeded
    at compileFileResources (file:///D:/crxjs-vite-dynamic-import/node_modules/@crxjs/vite-plugin/dist/index.mjs:1724:28)
    at compileFileResources (file:///D:/crxjs-vite-dynamic-import/node_modules/@crxjs/vite-plugin/dist/index.mjs:1724:7)
    at compileFileResources (file:///D:/crxjs-vite-dynamic-import/node_modules/@crxjs/vite-plugin/dist/index.mjs:1724:7)
    at compileFileResources (file:///D:/crxjs-vite-dynamic-import/node_modules/@crxjs/vite-plugin/dist/index.mjs:1724:7)
    at compileFileResources (file:///D:/crxjs-vite-dynamic-import/node_modules/@crxjs/vite-plugin/dist/index.mjs:1724:7)
    at compileFileResources (file:///D:/crxjs-vite-dynamic-import/node_modules/@crxjs/vite-plugin/dist/index.mjs:1724:7)
    at compileFileResources (file:///D:/crxjs-vite-dynamic-import/node_modules/@crxjs/vite-plugin/dist/index.mjs:1724:7)
    at compileFileResources (file:///D:/crxjs-vite-dynamic-import/node_modules/@crxjs/vite-plugin/dist/index.mjs:1724:7)
    at compileFileResources (file:///D:/crxjs-vite-dynamic-import/node_modules/@crxjs/vite-plugin/dist/index.mjs:1724:7)
    at compileFileResources (file:///D:/crxjs-vite-dynamic-import/node_modules/@crxjs/vite-plugin/dist/index.mjs:1724:7)
[crx:manifest-post] Error in crx:web-accessible-resources.renderCrxManifest
✓ built in 122ms
error during build:
[...]

PR #720, opened 2023-05-26
fix: infinite recursion on circular dependency

fractalo:
In order to handle the case of a circular dependency on js chunk, 
`compileFileResources` function has been modified to return immediately in the case of processed files.

[#719](https://github.com/crxjs/chrome-extension-tools/issues/719)


jacksteamdev (approved):
@fractalo This looks good to me, thanks for the PR! Technically, we should add tests for these, but this time I'll just pull the repro from the issue into the output test folder.

Issue #695, opened 2023-04-25
CRXJS doesn't work with content script in "MAIN" world


svicalifornia:
### Build tool

Vite

### Where do you see the problem?

- [X] In the browser
- [ ] In the terminal

### Describe the bug

When my manifest (V3) has a content script with `"world": "MAIN"`, the following error appears in the Chrome console:

```
TypeError: Cannot read properties of undefined (reading 'getURL')
    at my_content_script.js-loader.js:13:22
```

In the Sources tab, I see that my_content_script.js-loader.js contains:

```js
(function () {
  'use strict';

  const injectTime = performance.now();
  (async () => {
    if ("")
      await import(
        /* @vite-ignore */
        chrome.runtime.getURL("")
      );
    await import(
      /* @vite-ignore */
      chrome.runtime.getURL("vendor/vite-client.js")
    );
    const { onExecute } = await import(
      /* @vite-ignore */
      chrome.runtime.getURL("src/my_content_script.js")
    );
    onExecute?.({ perf: { injectTime, loadTime: performance.now() - injectTime } });
  })().catch(console.error);

})();
```

CRXJS (or Vite?) is trying to call `chrome.runtime.getURL` to get the JS paths to import. However, `chrome.runtime` is not defined in the MAIN world.

These imports should be rewritten to first check for the existence of `chrome.runtime` — if it doesn't exist, then some sort of shim should be defined for `chrome.runtime.getURL` to get the desired paths.

As one way to do this, CRXJS could:
* Automatically add another content script in the isolated world
* Establish bidirectional communication between the two worlds (via events on `document.documentElement`)
* Send a `CRXJS_getRootURL` event to from MAIN to the isolated world
* Handle the event in the isolated world by calling `chrome.runtime.getURL('')` and sending it back to MAIN
* Awaiting that response in the MAIN, then using it to define a shim for `chrome.runtime.getURL`, then continuing with the imports as shown in the code block above.
  

### Reproduction

mainfest.json:
```json
  "content_scripts": [
    {
      "matches": ["<all_urls>"],
      "js": ["src/my_content_script.js"],
      "run_at": "document_start",
      "world": "MAIN"
    }
  ]
```

### Logs

```shell
TypeError: Cannot read properties of undefined (reading 'getURL')
    at my_content_script.js-loader.js:13:22
```

### System Info

```shell
System:
    OS: macOS 12.6.5
    CPU: (8) arm64 Apple M1
[...]

PR #744, opened 2023-07-01
support content_scripts[i].world in manifest

### LINKED ISSUE #741: MV3 Schema is missing world property for content scripts
### Build tool

Rollup

- [ ] In the browser
- [X] In the terminal

### Describe the bug

https://github.com/crxjs/chrome-extension-tools/blob/5bf6711ec75e74f31b22474a7465a72a052089e0/schema/manifest-v3.schema.json#L272
In the schema the property "world" is missing which is in the mv3 docs: https://developer.chrome.com/docs/extensions/mv3/content_scripts/#functionality

This property is pretty new: https://bugs.chromium.org/p/chromium/issues/detail?id=1330986

Currently when I use the property, it creates an error that an additional property exists and prevents any further execution. 

### Reproduction

.

### Logs

```shell
.
```

### System Info

```shell
.
```

### Severity

blocking all usage of RPCE

TheColaber:
Resolves #741 

Without this change it will error in the console which blocks all usage of rpce


Toumash:
In vite i can add `//@ts-ignore`and it starts the dev server fine (and adds it to the resulting manifest.json)

Unfortunately it does not matter since it cannot be loaded in the browser itself since mainworld scripts do not support `chrome.runtime` apis on which we depend for loading modules. Thats why jack [blogged about it here with instruction](https://dev.to/jacksteamdev/advanced-config-for-rpce-3966#main-world-scripts).
Theres also whole thread about it here #695.
Does it work for you in rollup?


> TypeError: Cannot read properties of undefined (reading 'getURL')
    at index.tsx-loader.js:9:24
    at index.tsx-loader.js:20:5
    at index.tsx-loader.js:22:3
    
<details>
<summary>Screenshots</summary>

image

image

</details>

jacksteamdev:
I'd like for CRXJS to support this feature, but there's more work to be done before this PR will work. For now, I'm going to decline merging this b/c I don't want to create extra confusion.

## Why Not?

Vite is an ES modules build tool, but Chrome Extension content scripts are plain scripts, not ESM. We can support ESM in content scripts using `import()` and this is what CRXJS does. It makes a loader file that looks like this:

```typescript
(function () {
  'use strict';

  const injectTime = performance.now();
  (async () => {
    if ("")
      await import(
        /* @vite-ignore */
        chrome.runtime.getURL("")
      );
    await import(
      /* @vite-ignore */
      chrome.runtime.getURL("vendor/vite-client.js")
    );
    const { onExecute } = await import(
      /* @vite-ignore */
      chrome.runtime.getURL("src/content.ts.js")
    );
    onExecute?.({ perf: { injectTime, loadTime: performance.now() - injectTime } });
  })().catch(console.error);

})();
```

Notice the use of `chrome.runtime.getURL` to get the right URL for the content script. This method isn't available in the main world, so we can't just add the `world: MAIN` field, we need to solve the problem of creating a non-ESM content script file.

## What Else

We need to find a way to use Vite to build an IIFE script. I'm sure this can be done, but I haven't been able to put in the hours to get it working.

PRs are appreciated.

PR #751, opened 2023-07-11
fix: Cannot read properties of undefined (reading 'scripts') at auditor

cutsin:


jacksteamdev (commented):
This looks fine to me. Thanks for the PR!

Would you mind providing a minimal reproduction of how this error might happen?

Arturo-Lopez:
> This looks fine to me. Thanks for the PR!
> 
> Would you mind providing a minimal reproduction of how this error might happen?

Just as a note, this issue happened to me when I use the latest beta version (beta-23) with `vite-plugin-html`. The solution was to remove `vite-plugin-html` from the plugin list. This is a sample project with the issue https://github.com/Arturo-Lopez/vite-extension-error

PR #753, opened 2023-07-16
Add explanation about Content Security Policy when adding a background script

andresgutgon:
# What?
I noticed that loading a background script failed until I add the CSP for `connect-src` because I think Vite server needs for connecting with the dev server websockets

### Issue on Vite
Here one person suggest that solution. 
https://github.com/vitejs/vite/discussions/7590

And it worked in my case. At least now my extension background loads without error : )


jacksteamdev:
This is interesting! I don't have this error in my projects, but I'm glad you found a solution that works for you 😸

farzadmf:
@andresgutgon was wondering if you have an example repo working for Firefox (as I'm guessing the CSP value is because of Firefox).

I have the same value as you have, but I'm seeing all these errors:
image

Not sure where that `'self' 'wasm-unsafe-eval'` is coming from as I clearly have this (I'm even directly checking out the generated `dist/manifest.json`):

```jsonc
{
  // ... other things
  "content_security_policy": {
    "extension_pages": "default-src 'self'; connect-src ws://localhost:5173;"
  },
  // ... other things
}
```

andresgutgon:
> I was wondering if you have an example repo working for Firefox 

I never tried with Firefox. I thought Firefox extensions were not the same. But as I said I never did one.

farzadmf:
Ah OK, cool, thank you for letting me know. Given that things work "as is" in Chrome (at least for me), and it's Firefox creating all the headaches with CSP, I thought your change is because of that as well

jacksteamdev:
Thank you for your contribution to CRXJS! I'm closing this PR since it's been open for more than 90 days without activity.

PR #759, opened 2023-07-25
plugin output should regard config.root

intellild:


packages/vite-plugin/src/node/fileWriter.ts
```diff
@@ -1,4 +1,5 @@
 import fsx from 'fs-extra'
+import path from 'path'
```
jacksteamdev, line 2: @intellild Have you tested this on Windows? IIRC, we need to use `node:path/posix` here for Windows compatibility.

jacksteamdev (changes requested):
The fix itself looks good to me, thanks for the contribution. We can merge after the following are completed.

## Tasks before merging
- [ ] Verify that it works on Windows
- [ ] Add an e2e test

intellild:
Sorry for the late, migrated to posix path

jacksteamdev:
Thank you for your contribution to CRXJS! I'm closing this PR since it's been open for more than 90 days without completing the requested changes (add e2e test).

PR #771, opened 2023-08-23
docs: add missing extension install guide for react

Toumash:
as [reported in the discussion here](https://github.com/crxjs/chrome-extension-tools/discussions/709) - docs section "Development Basics with React" is missing the extension installation part. It exist for other libraries

Before:
<details>

image

</details>

After:

<details>

image

</details>


AmySteam (approved):
Thanks @Toumash!! 💯 😃

PR #776, opened 2023-09-03
feat: add compatibility mode for Firefox

winstxnhdw:
See #644.

The original author seems to be inactive.

PR #790, opened 2023-09-25
fix: background scripts for firefox build

SaltyCucumber:
This is a follow up to https://github.com/crxjs/chrome-extension-tools/pull/776
I tried it out and the firefox build ignores background scripts, this PR fixes it

### DIFF
--- a/.changeset/fluffy-geckos-flash.md
+++ b/.changeset/fluffy-geckos-flash.md
@@ -1,0 +1,5 @@
+---
+"@crxjs/vite-plugin": patch
+---
+
+fix: background scripts for firefox build
--- a/packages/vite-plugin/src/node/plugin-background.ts
+++ b/packages/vite-plugin/src/node/plugin-background.ts
@@ -108,6 +108,7 @@ export const pluginBackground: CrxPluginFn = () => {
         } else {
           manifest.background = {
             scripts: [this.getFileName(refId)],
+            type: 'module',
           }
         }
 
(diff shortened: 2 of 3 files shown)


Chimildic:
@jacksteamdev can you add the request to npm version?

fbjorn:
Hi @SaltyCucumber , can this PR by any chance fix #801 ? And have you experienced this problem with Firefox at all?
Seems like this plugin generates top level `service-worker-loader.js` that imports other files, but Firefox doesn't support imports. 
There's even a comment about that: https://github.com/crxjs/chrome-extension-tools/blob/main/packages/vite-plugin/src/node/plugin-background.ts#L69
But as far as I can see, the line above only affects development mode, not the built files

SaltyCucumber:
> Hi @SaltyCucumber , can this PR by any chance fix #801 ? And have you experienced this problem with Firefox at all? Seems like this plugin generates top level `service-worker-loader.js` that imports other files, but Firefox doesn't support imports. There's even a comment about that: https://github.com/crxjs/chrome-extension-tools/blob/main/packages/vite-plugin/src/node/plugin-background.ts#L69 But as far as I can see, the line above only affects development mode, not the built files

Are you trying to use a dev build with HMR?
Cause the code I'm updating in this PR only works on prod build

Stevenkhalife:
@jacksteamdev  This PR fixes some issues in #776 where the background script was being completely ignored on build. Can we merge and deploy it asap so that we can benefit from the newly introduced Firefox support?

jacksteamdev:
@SaltyCucumber @Stevenkhalife I've done a code review but haven't had time to test this live. If there are any bugs, it's up to the community to make a PR. Thanks for the contributions!

Issue #618, opened 2022-12-26
Dev client worker capturing /_favicon/ request in chrome manifest v3


stevezhu:
### Build tool

Vite

### Where do you see the problem?

- [X] In the browser
- [ ] In the terminal

### Describe the bug

I'm attempting to use the `/_favicon/` api from chrome manifest v3 shown here:
https://github.com/GoogleChrome/chrome-extensions-samples/tree/main/api/favicon

It seems that due to the [hmr client worker](https://github.com/crxjs/chrome-extension-tools/blob/main/packages/vite-plugin/src/client/es/hmr-client-worker.ts) capturing all requests during dev, the requests to `/_favicon/` are also being captured instead of being handled by chrome.

### Reproduction

Use https://github.com/GoogleChrome/chrome-extensions-samples/tree/main/api/favicon with `@crxjs/vite-plugin` in dev mode.

### Logs

Failed request (when client worker is capturing requests):
```
Request URL: http://localhost:5173/_favicon/?pageUrl=https%3A%2F%2Fwww.google.com%2F&size=64&t=1672027377433
Request Method: GET
Status Code: 404 Not Found
Remote Address: 127.0.0.1:5173
Referrer Policy: strict-origin-when-cross-origin
```

Expected successful request:
```
Request URL: chrome-extension://<id>/_favicon/?pageUrl=https%3A%2F%2Fwww.google.com%2F&size=64
Request Method: GET
Status Code: 200 OK (from memory cache)
Referrer Policy: strict-origin-when-cross-origin
```

### System Info

```shell
System:
    OS: macOS 13.0.1
    CPU: (8) x64 Intel(R) Core(TM) i5-1038NG7 CPU @ 2.00GHz
    Memory: 453.57 MB / 16.00 GB
    Shell: 3.5.1 - /usr/local/bin/fish
  Binaries:
    Node: 18.8.0 - /usr/local/bin/node
    Yarn: 1.22.19 - /usr/local/bin/yarn
    npm: 8.18.0 - /usr/local/bin/npm
  npmPackages:
    @crxjs/vite-plugin: 2.0.0-beta.9 => 2.0.0-beta.9
    vite: ^4.0.1 => 4.0.1
```

### Severity

annoyance

zhaoguangyue:
When is it expected to be repaired

mj2ks:
@stevezhu Found a solution?

alajmo:
[...]

PR #819, opened 2023-11-02
fix: ignore requests to /_favicon/

alajmo:
Fixes https://github.com/crxjs/chrome-extension-tools/issues/618

### DIFF
--- a/.changeset/pre.json
+++ b/.changeset/pre.json
@@ -22,6 +22,7 @@
     "fluffy-numbers-brake",
     "fresh-suns-do",
     "gold-spies-occur",
+    "great-fishes-remember",
     "grumpy-brooms-exist",
     "grumpy-comics-explain",
     "happy-carpets-boil",
--- a/packages/vite-plugin/CHANGELOG.md
+++ b/packages/vite-plugin/CHANGELOG.md
@@ -1,5 +1,11 @@
 # @crxjs/vite-plugin
 
+## 2.0.0-beta.32
+
+### Patch Changes
+
+- 8525e04: Propagate headers in hmr-client-worker.ts/sendToServer
+
 ## 2.0.0-beta.31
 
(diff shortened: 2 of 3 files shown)


packages/vite-plugin/src/client/es/hmr-client-worker.ts
```diff
@@ -34,6 +34,15 @@ self.addEventListener('fetch', (fetchEvent) => {
...
  */
 async function sendToServer(url: URL): Promise<Response> {
+	if (url.href.includes('/_favicon/')) {
+		const response = await fetch(url.href)
+		return new Response(response.body, {
+			headers: {
+				'Content-Type': response.headers.get('Content-Type') ?? 'text/javascript',
+			},
+		})
+	}
```
jacksteamdev, line 44: I imagine there are other use cases to not redirect to the dev server. Let's move `url.href.includes('/_favicon/')` up into line 23 and abstract it into a guard function.

jacksteamdev (changes requested):
Thanks for figuring this out! 💯 I requested a small change.

packages/vite-plugin/src/client/es/hmr-client-worker.ts
alajmo: Agree, I put them in a array `const ignoreURLPaths = ['/_favicon/']`, in case there are other paths to ignore as well (also perhaps allow users to add their own ignore paths). It's not as efficient as doing separate if statements as before (though V8 is probably optimizing these on its own), but as I mentioned in the future users might want to add other ignore paths?

jacksteamdev:
This PR looks good, but it needs to be reviewed and tested by a member of the CRXJS maintenance team before merging.

Toumash:
it works here: https://github.com/crxjs/chrome-extension-tools/pull/1028

PR #833, opened 2023-11-20
fix: hmr error

ZainChen:
# Problems with `hmr` not updating files referenced by both `background` and other page components

Reason: For example, the `dist/src/constants/message.ts.js` file, which is referenced by `background` and other components, `return [];` prevents all of the default `hmr` behavior of `handleHotUpdate`. The default `hmr` cannot be blocked.


jacksteamdev (commented):
Nice catch ⚾ Thanks for the help!

PR #834, opened 2023-11-21
removing unnecessary v8 dependency

tstachl:
This fixes an issue with deno. Deno doesn't support `v8.serialize` and `v8.deserialize`, however both Deno and node have a native `structuredClone` method. Hence I removed the helper method.

Now it works with both node and deno.


jacksteamdev (approved):
Looking good! `structuredClone` has been available since Node 17, and LTS is currently 18, so I think we can merge this.

jacksteamdev (changes requested):
Sorry, I spoke too soon. 😓 This isn't passing CI, so we need to update the tsconfig and fix the tests to include `structuredClone`.

PR #868, opened 2024-02-21
Update 02-background.md

heymaaz:
Add details of what a service worker is and why would you need it? Give an example of src/background.ts


Toumash:
Im not sure what this PR is supposed to be, but # is not a comment in markdown and it looks like you are trying to add a todo instead of adding the explanation yourself. 
A PR should contain a solution not a task - if you want to start a discussion use a https://github.com/crxjs/chrome-extension-tools/discussions github page

jacksteamdev:
@Toumash thank you!

@heymaaz Contributions are welcome! There's a link at the bottom of the doc to the Google Dev Docs for Chrome Extension service workers.

PR #870, opened 2024-02-27
docs: added Vue use cases

xianjianlf2:
I found that the documentation lacks information on how to use `content-script` of `vue`.

### DIFF
--- a/packages/vite-plugin-docs/docs/getting-started/vue/02-add-content-script.md
+++ b/packages/vite-plugin-docs/docs/getting-started/vue/02-add-content-script.md
@@ -1,0 +1,121 @@
+---
+id: add-content-script
+title: Add a content script
+description: Add a Vue content script to an existing project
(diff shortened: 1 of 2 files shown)


Toumash:
Looks like its ready to be merged. Just a copy of the page from react + vue code.

xianjianlf2:
The document lacks an example of a `content script` using `vue` syntax, similar to the [react content script](https://crxjs.dev/vite-plugin/getting-started/react/add-content-script). Here is the relevant content added. Since the syntax of `react 18` is similar to that of `vue`, but in Vue, examples are written on `SFC` components.

> Looks like its ready to be merged. Just a copy of the page from react + vue code.

jacksteamdev (approved):
@xianjianlf2 Thanks 🙏

jacksteamdev:
@Toumash thanks for the review!

PR #915, opened 2024-09-04
feat: add doc search

hichemfantar:
image


jacksteamdev:
@hichemfantar This is a great addition! Thanks for the hard work.

PR #916, opened 2024-09-09
fix background.type manifest property typescript type

### LINKED ISSUE #914: manifest type error: Type 'string' is not assignable to type '"module"'
### Build tool

Vite

- [ ] In the browser
- [ ] In the terminal

### Describe the bug

[Documentation ](https://crxjs.dev/vite-plugin/concepts/background) says we should include type:"module" to the background service worker but the types don't like this.

image

image

### Reproduction

https://crxjs.dev/vite-plugin/concepts/background

### Logs

_No response_

### System Info

```shell
  System:
    OS: Windows 10 10.0.19045
    CPU: (12) x64 Intel(R) Core(TM) i5-10500H CPU @ 2.50GHz
    Memory: 5.42 GB / 15.78 GB
  Binaries:
    Node: 18.18.2 - C:\Program Files\nodejs\node.EXE
    Yarn: 1.22.21 - C:\Program Files\nodejs\yarn.CMD
    npm: 9.8.1 - C:\Program Files\nodejs\npm.CMD
    pnpm: 9.5.0 - C:\Program Files\nodejs\pnpm.CMD
  Browsers:
    Edge: Chromium (127.0.2651.74)
    Internet Explorer: 11.0.19041.4355
  npmPackages:
    @crxjs/vite-plugin: ^1.0.14 => 1.0.14
    vite: ^2.9.9 => 2.9.18
```

### Severity

annoyance

hichemfantar:
this fixes the issue using the manifest.json schema instead of the config file https://json.schemastore.org/chrome-manifest.json
image

The issue arises because the schema you're trying to convert indicates that the type is a string but only allows a specific value ("module"). If you're using this TypeScript type for something that can be either a specific string or a more general string type, you may want to handle it differently.

If you want IntelliSense to work and ensure that "module" is a valid and suggested value, while still allowing other strings, you can use string literal unions in a way that provides IntelliSense but also accepts any string.
Using a branded type approach (to enforce "module" while still allowing any string but guiding IntelliSense):
`type Type = "module" | (string & {});`

closes #914


jacksteamdev (approved):
LGTM

PR #923, opened 2024-09-30
Proxy Cache-Control header

rogueg:
I was running into an issue where refreshing an extension page would load stale code.

This happened because without the cache-control header, refreshing would use the cached version of index.html. Vite already sets the header appropriately, so we just need to pass it down through the service worker.


jacksteamdev (approved):
@rogueg Fantastic! Thanks for fixing this

jacksteamdev:
@rogueg This landed in `@crxjs/vite-plugin@2.0.0-beta.28`

PR #925, opened 2024-10-11
Fix: Adding correct projectType for VitePlugin docs

Priyanshu-001:
Minor fix for current docs site says to create vanilla js project even for react and vue based extensions. eg [here](https://crxjs.dev/vite-plugin/getting-started/vue/create-project). I think it should say react/vue there instead of vanilla js.  so i did a minor fix to correct those. 

this required refactoring of  _create-project-tabs.mdx to a jsx style so the proejctType could be parameterized The new versions looks similar let me know if you think this might cause any issue.  

ss for new versions
image

thanks


jacksteamdev (approved):
LGTM! Thanks for the PR!

Issue #946, opened 2024-10-27
How to disable content-script loader


mp3por:
### Build tool

Rollup

### Where do you see the problem?

- [ ] In the browser
- [ ] In the terminal

### Describe the bug

At the moment when I run `vite build` I get a `content-script.ts-loader.js` file which looks something like this:

```
(function () {
  'use strict';

  const injectTime = performance.now();
  (async () => {
    const { onExecute } = await import(
      /* @vite-ignore */
      chrome.runtime.getURL("assets/content-script.ts-C7vMt-6J.js")
    );
    onExecute?.({ perf: { injectTime, loadTime: performance.now() - injectTime } });
  })().catch(console.error);

})();
```

Because of this I receive the following error in the browser

```
content-script.ts-loader-D9s8tIAp.js:8 Refused to load the script 'chrome-extension://300fed4b-9168-4f91-9e64-9e3c97c6aa31/assets/content-script.ts-C7vMt-6J.js' because it violates the following Content Security Policy directive: "script-src 'self' 'wasm-unsafe-eval' 'inline-speculation-rules'". Note that 'script-src-elem' was not explicitly set, so 'script-src' is used as a fallback.
```

I use the following content_script_policy:

```
content_security_policy: {
    extension_pages: "script-src 'self' 'wasm-unsafe-eval'; object-src 'self';",
  },
```

How to tell the plugin to not add this `loader` file because i don't need it in production ?

### Reproduction

asdjioasd

### System Info

```shell
System:
    OS: macOS 15.0.1
    CPU: (10) arm64 Apple M1 Max
    Memory: 71.64 MB / 32.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 20.12.2 - ~/.nvm/versions/node/v20.12.2/bin/node
    npm: 10.5.0 - ~/.nvm/versions/node/v20.12.2/bin/npm
    pnpm: 9.11.0 - /opt/homebrew/bin/pnpm
    Watchman: 2024.09.23.00 - /opt/homebrew/bin/watchman
  Browsers:
    Brave Browser: 130.1.71.118
    Chrome: 130.0.6723.70
    Safari: 18.0.1
  npmPackages:
    @crxjs/vite-plugin: ^2.0.0-beta.28 => 2.0.0-beta.28
    vite: ^5.4.1 => 5.4.2
```

### Severity

blocking all usage of RPCE

noahajohn:
[...]

PR #997, opened 2025-03-14
fix: do not use IIFE for contents scripts without imports

refi93:
Motivation: wrapping in IIFE may delay script injection, which is a needless tradeoff if the script has no (ES module) imports

inspired by https://github.com/samrum/vite-plugin-web-extension/blob/main/src/utils/loader.ts#L57-L66

### DIFF
--- a/packages/vite-plugin/src/node/plugin-contentScripts.ts
+++ b/packages/vite-plugin/src/node/plugin-contentScripts.ts
@@ -135,7 +135,7 @@ export const pluginContentScripts: CrxPluginFn = () => {
           },
         }
       },
-      generateBundle() {
+      generateBundle(_options, bundle) {
         // emit content script loaders
         for (const [key, script] of contentScripts)
           if (key === script.refId) {
@@ -145,12 +145,28 @@ export const pluginContentScripts: CrxPluginFn = () => {
             } else if (script.type === 'loader') {
               const fileName = this.getFileName(script.refId)
               script.fileName = fileName
-              const refId = this.emitFile({
-                type: 'asset',
-                name: getFileName({ type: 'loader', id: basename(script.id) }),
-                source: createProLoader({ fileName }),
-              })
-              script.loaderName = this.getFileName(refId)
+
+              const bundleFileInfo = bundle[fileName]
+              // the loader uses IIFE which in this case needlessly
+              // delays content script execution which may not be desired
+              const shouldUseLoader = !(
+                bundleFileInfo.type === 'chunk' &&
+                bundleFileInfo.imports.length === 0 &&
+                bundleFileInfo.dynamicImports.length === 0
+              )
+
+              if (shouldUseLoader) {
+                const refId = this.emitFile({
+                  type: 'asset',
+                  name: getFileName({
+                    type: 'loader',
+                    id: basename(script.id),
+                  }),
+                  source: createProLoader({ fileName }),
+                })
+
+                script.loaderName = this.getFileName(refId)
+              }
             } else if (script.type === 'iife') {
               throw new Error('IIFE content scripts are not implemented')
             }


Toumash:
related: https://github.com/crxjs/chrome-extension-tools/issues/946#issuecomment-2450057235

packages/vite-plugin/src/node/plugin-contentScripts.ts
```diff
@@ -167,6 +167,8 @@ export const pluginContentScripts: CrxPluginFn = () => {
                 })
 
                 script.loaderName = this.getFileName(refId)
+              } else {
+                bundleFileInfo.code = `(function(){${bundleFileInfo.code}})()`
```
refi93, line 171: We realized that the original loader which wrapped the content script in a function call had the side effect of variable scope isolation, which, if droppped, may result in a breaking change, if some variable names across different content scripts are the same (especially after vite mangles variable names). To preserve this behavior, we are wrapping here the content script in a (synchronous) self invoked function which should cause no issues since the script has no imports/exports

Toumash: now the comment is unclear
> // the loader uses IIFE which in this case needlessly
              // delays content script execution which may not be desired

because now the solution uses IIFE itself 
> // make sure the code is wrapped in a function invocation
   // to have the same scope isolation as the loader provides

``(function(){${bundleFileInfo.code}})()`` is a IIFE

refi93: thanks, fixed 👍

Toumash (changes requested):
lgtm, still needs to be run locally to test. 
❌No changeset file is present, so the e2e tests didnt run here yet

refi93:
> No changeset file is present, so the e2e tests didnt run here yet

fixed 👍 https://github.com/crxjs/chrome-extension-tools/pull/997/commits/794b97daca8e419239a4a1f84a03fe8901e5508b

packages/vite-plugin/tests/e2e/mv3-vite-svelte-content-script/__image_snapshots__/vite-build-0.png
refi93, line 1: this change looks suspicious, however I'm getting the same diff also when I run `pnpm test:update` on `main` so I guess this change is not related to this PR, or perhaps I have something misconfigured locally

I guess https://github.com/crxjs/chrome-extension-tools/pull/999 fixes that

packages/vite-plugin/tests/out/vite-svelte/__snapshots__/serve.test.ts.snap
```diff
@@ -365,7 +365,7 @@ exports[`serve fs output > src/content.js-loader.js 1`] = `
 `;
 
 exports[`serve fs output > src/lib/Counter.svelte.js 1`] = `
-"import { createHotContext as __vite__createHotContext } from \\"/vendor/vite-client.js\\";import.meta.hot = __vite__createHotContext(\\"/src/lib/Counter.svelte.js\\");/* src/lib/Counter.svelte generated by Svelte v3.48.0 */
+"import { createHotContext as __vite__createHotContext } from \\"/vendor/vite-client.js\\";import.meta.hot = __vite__createHotContext(\\"/src/lib/Counter.svelte.js\\");/* src/lib/Counter.svelte generated by Svelte v3.59.2 */
```
Toumash, line 368: how is that different? no `package-lock.json` or not running `npm ci`?

packages/vite-plugin/tests/out/with-sourcemaps/__snapshots__/build.test.ts.snap
```diff
@@ -41,49 +41,31 @@ Object {
...
-  (async () => {
-    const { onExecute } = await import(
-      /* @vite-ignore */
-      chrome.runtime.getURL(\\"assets/content.ts.hash1.js\\")
-    );
-    onExecute?.({ perf: { injectTime, loadTime: performance.now() - injectTime } });
-  })().catch(console.error);
-
-})();
+})()//# sourceMappingURL=data:application/json;charset=utf-8;base64,eyJ2ZXJzaW9uIjozLCJmaWxlIjoiY29udGVudC50cy5oYXNoNWFmYjk4MDAuanMiLCJzb3VyY2VzIjpbIi4uLy4uL3NyYy9jb250ZW50LnRzIl0sInNvdXJjZXNDb250ZW50IjpbInR5cGUgTG9nID0gJ2NvbnRlbnQgc2NyaXB0J1xuY29uc3QgbWVzc2FnZTogTG9nID0gJ2NvbnRlbnQgc2NyaXB0J1xuY29uc29sZS5sb2cobWVzc2FnZSlcbmV4cG9ydCB7fVxuIl0sIm5hbWVzIjpbXSwibWFwcGluZ3MiOiJBQUNBLE1BQU0sVUFBZTtBQUNyQixRQUFRLElBQUksT0FBTzsifQ==
```
Toumash, line 63: how is that different. Looks like outdated tests on main

packages/vite-plugin/tests/out/vite-react-fast-refresh/__snapshots__/serve.test.ts.snap
```diff
@@ -170,15 +170,42 @@ export default App;
 var _c;
 $RefreshReg$(_c, \\"App\\");
 if (import.meta.hot) {
+  let isReactRefreshBoundary = function(mod) {
```
Toumash, line 173: some updated module? How is this different?

packages/vite-plugin/tests/out/vite-content-script-css-imports-2/__snapshots__/build.test.ts.snap
```diff
@@ -73,48 +71,14 @@ exports[`build fs output > assets/index.hash3.css 1`] = `
...
-"(function () {
-  'use strict';
-
-  const injectTime = performance.now();
-  (async () => {
-    const { onExecute } = await import(
-      /* @vite-ignore */
-      chrome.runtime.getURL(\\"assets/index.ts.hash4.js\\")
-    );
-    onExecute?.({ perf: { injectTime, loadTime: performance.now() - injectTime } });
```
Toumash, line 115: - do we need some kind of performance metrics for this scenario?

packages/vite-plugin/tests/e2e/mv3-vite-svelte-content-script/__image_snapshots__/vite-build-0.png
Toumash: Yes, it is probably why it was changed in mentioned PR - because google.com added a cookie policy that covers the page on load preventing the test from running normally

elwin013: Yes, that's correct - I've updated the description in the PR.

google.com adding cookie consent is the one reason, second is instability of the test due to broken image - sometimes the screenshot is done correctly after showing "broken image 'icon'", but more commonly when it is not loaded yet. Thus image is removed :slightly_smiling_face:

packages/vite-plugin/tests/out/vite-content-script-css-imports-2/__snapshots__/build.test.ts.snap
refi93: good catch - indeed, looks like `onExecute` is an optional callback that content scripts may export which the crxjs loader should call: https://github.com/crxjs/chrome-extension-tools/pull/689

given that in order to call a content script directly we require it to have no imports/exports, it is by definition not needed to try to call such callback in case of "direct" injection. I will add a comment clarifying that for future reference 👍

refi93: https://github.com/crxjs/chrome-extension-tools/pull/997/commits/f6755ccbcb6b7a4223f75f97687cff365e128a3d

packages/vite-plugin/tests/out/vite-react-fast-refresh/__snapshots__/serve.test.ts.snap
refi93: good catch - turns out I had an outdated version of pnpm which couldn't handle the pnpm-lock.yaml file format, which resulted in pnpm ignoring it and ultimately in mismatched dependencies in my local setup, which caused unrelated test runs to diverge like this. Sorry for the confusion, should be fixed with latest force-push

packages/vite-plugin/tests/out/with-sourcemaps/__snapshots__/build.test.ts.snap
refi93: same issue as https://github.com/crxjs/chrome-extension-tools/pull/997#discussion_r2030621822 - fixed with latest force-push

packages/vite-plugin/tests/out/vite-svelte/__snapshots__/serve.test.ts.snap
refi93: same issue as https://github.com/crxjs/chrome-extension-tools/pull/997#discussion_r2030621822 - fixed with latest force-push

packages/vite-plugin/tests/e2e/mv3-vite-svelte-content-script/__image_snapshots__/vite-build-0.png
Toumash, line 1: i guess we need to either merge the image fix to the `main` first or update the snaptshots in this PR

Toumash (changes requested):
👍 IMO ok
but not ready to be merged yet tho
we need to either: 
- wait for the https://github.com/crxjs/chrome-extension-tools/pull/999 to merge to have the image snapshot tests resolved 
- apply the same fixes in this PR
- ➡️create a separate PR that resolves the image snapshots first (so we'll not block any other PRs and will be super easy to merge)

Lets create a separate PR https://github.com/crxjs/chrome-extension-tools/pull/1001 and wait for it to merge

Toumash (approved):

refi93:
@Toumash just curious - is there anything preventing this from being merged? If so, can I somehow help with it?

colearendt:
Looks great! Thanks for your patience here!

salmin89:
I noticed that this PR only solved this for generated builds, but in dev mode the content scripts still use loaders. Is this because HMR wouldn't work or something that was overlooked?

Toumash:
> I noticed that this PR only solved this for generated builds, but in dev mode the content scripts still use loaders. Is this because HMR wouldn't work or something that was overlooked?

quick research with opencode:


## Answer
**This is intentional, not an oversight.** The loader is required in dev mode for HMR to work.
### Why Dev Mode Requires the Loader
In dev mode, the content script loader does three critical things:
1. **Imports the Vite client** (`/@vite/client`) — essential for HMR support
2. **Imports the React preamble** — for React Fast Refresh when using React
3. **Uses dynamic `import()`** — loads the actual content script after HMR infrastructure is ready
### Code Evidence
In `packages/vite-plugin/src/node/plugin-contentScripts.ts`, the dev mode handler (`apply: 'serve'`) always injects the Vite client:
```typescript
const client = add({ type: 'module', id: viteClientId })
// ...
source: createDevLoader({
  preamble: preamble.fileName,
  client: client.fileName,   // <-- HMR client is ALWAYS injected
  fileName: file.fileName,
}),
```
The dev loader in `packages/vite-plugin/src/client/iife/content-dev-loader.ts`:
```typescript
await import(/* @vite-ignore */ chrome.runtime.getURL(__PREAMBLE__))
await import(/* @vite-ignore */ chrome.runtime.getURL(__CLIENT__))  // HMR client
const { onExecute } = await import(/* @vite-ignore */ chrome.runtime.getURL(__SCRIPT__))
```
### Why the Optimization Cannot Apply to Dev Mode
| Mode | Loader Needed? | Reason |
|------|----------------|--------|
| **Production** | Only if has imports | No HMR needed, scripts can run synchronously |
| **Dev** | Always | Must inject Vite HMR client before script runs |
- Even the simplest content script (`console.log("hello")`) needs the Vite client loaded first for HMR
- Chrome's content script injection doesn't support ES modules with dependencies
- The async loader is the only way to inject the HMR client before the content script executes
---
**Bottom line:** The loader in dev mode isn't just for handling content script imports — it's architecturally required to bootstrap the Vite HMR infrastructure before the content script runs.

PR #998, opened 2025-03-28
Bug/back forward cache

salmin89:
This PR is to silence this error:
> Unchecked runtime.lastError: The page keeping the extension port is moved into back/forward cache, so the message channel is closed.

Happens when you are using ports and navigate away / back.


colearendt (approved):
This looks great - thanks @salmin89 !

PR #1014, opened 2025-06-06
[WIP]feat: add TypeScript type checking for manifest file paths

FliPPeDround:
This PR uses TypeScript types to enforce that the ManifestFilePath provided by the user must adhere to crxjs specifications: it cannot be an absolute or relative path, and the file must have an extension.
https://crxjs.netlify.app/concepts/manifest#manifest-paths

### error
```
./src/popup.html 
```
image

### right
```
src/popup.html 
```
image

### DIFF
--- a/packages/vite-plugin/src/node/defineManifest.ts
+++ b/packages/vite-plugin/src/node/defineManifest.ts
@@ -1,12 +1,16 @@
 import type { ConfigEnv } from 'vite'
 import type { ManifestV3, WebAccessibleResourceByMatch } from './manifest'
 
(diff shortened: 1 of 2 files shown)


Toumash:
maybe we could issue a warning in the terminal somehow? 
Why did you close it? ts wont work?

FliPPeDround:
1. Adding runtime checks for terminal warnings would slow down loading. I currently want only a simple compile-time error via TypeScript's type system, which users can also ignore with `@ts-ignore`. If later on everyone thinks runtime detection is necessary, I think it can be added.

2. The current implementation affects other types. We should implement path validation types specifically for the `defineManifest` method instead of modifying all `ManifestV3` types.

3. I'll reopen this PR after making the changes.

PR #1015, opened 2025-06-07
feat: add TypeScript type checking for manifest file paths

FliPPeDround:
This PR uses TypeScript types to enforce that the ManifestFilePath provided by the user must adhere to crxjs specifications: it cannot be an absolute or relative path, and the file must have an extension.
https://crxjs.netlify.app/concepts/manifest#manifest-paths

### error
```
./src/popup.html 
```
image

### right
```
src/popup.html 
```
image

# But
However, the current validation implementation prevents the JSON file from being properly validated when passed in. It might be possible to implement this, but I'm unsure how.
b0e6a7b5-730b-47ab-91df-638817855701

Alternatively, we could use [unconfig](https://github.com/antfu-collective/unconfig) to fetch the config file. This would provide validation support when users need to work with config files, without requiring explicit imports.

**This PR will remain unmerged until we reach a consensus in the discussion.**

### DIFF
--- a/packages/vite-plugin/src/node/defineManifest.ts
+++ b/packages/vite-plugin/src/node/defineManifest.ts
@@ -3,9 +3,6 @@ import type { FirefoxManifestBackground, ManifestV3, WebAccessibleResourceByMatc
 
 type Code = '.' | '/' | '\\'
 
-/**
- * aaaaaaaa
- */
 export type ManifestFilePath<T extends string> =
   T extends `${Code}${string}`
     ? never
--- a/packages/vite-plugin/src/node/defineManifest.ts
+++ b/packages/vite-plugin/src/node/defineManifest.ts
@@ -1,16 +1,74 @@
 import type { ConfigEnv } from 'vite'
-import type { ManifestV3, WebAccessibleResourceByMatch } from './manifest'
+import type { FirefoxManifestBackground, ManifestV3, WebAccessibleResourceByMatch } from './manifest'
 
-export type ManifestV3Export<T extends string> = ManifestV3<T> | Promise<ManifestV3<T>> | ManifestV3Fn<T>
+type Code = '.' | '/' | '\\'
 
-export type ManifestV3Fn<T extends string> = (env: ConfigEnv) => ManifestV3<T> | Promise<ManifestV3<T>>
+/**
+ * aaaaaaaa
+ */
+export type ManifestFilePath<T extends string> =
+  T extends `${Code}${string}`
+    ? never
+    : T extends `${string}.${infer Ext}`
+      ? Ext extends ''
+        ? never
+        : T
+      : never
 
-// export const defineManifest = (manifest: ManifestV3Export): ManifestV3Export =>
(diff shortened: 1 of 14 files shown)


Toumash (approved):
:+1:

after linter fix it is fine

FliPPeDround:
> 👍
> 
> after linter fix it is fine

It's not OK, the current validation implementation prevents the JSON file from being properly validated when passed in.

Toumash:
What about that runtime check? Any idea how to show a warning during `vite`?
Or maybe it is a topic for our own linter?

FliPPeDround:
> What about that runtime check? Any idea how to show a warning during `vite`? Or maybe it is a topic for our own linter?

If runtime detection is used, you can just throw an error directly in the plugin.
```ts
export default defineConfig({
  plugins: [
    {
      name: 'throw-error',
      buildStart() {
        throw new Error('This is an error')
      },
    }
  ],
})
```
However, I believe if this can be implemented using TypeScript type checking, it would surface issues earlier - catching errors during development itself.

Toumash:
Is jsdoc working right now? 
This should be the starting point to just add it as below.
It is good that the new crxjs docs website points the user into using `defineManifest` so that the user will read the jsdoc as he types instead of searching for something in the docs website or even worst as it is currently in the github issues

```tsx
/**
something
*/
default_popup: string;
```

FliPPeDround:
JSDoc certainly works fine, and I plan to add JSDoc as guidance. However, my current concern is: whether we should use TypeScript's type system to verify path accuracy. If we do, using JSON as the configuration file would cause type errors. I believe this would be a breaking change and should be carefully considered.

FliPPeDround:
I've found what I believe to be a relatively balanced approach: only when the user uses defineManifest, will the path field be validated to ensure it complies with the specifications. I've also added JSDoc comments to explain this.
@Toumash cc

Toumash:
hmm interesting. `defineManifest` doesnt do anything important?

update: yes, the whole source code for `defineManifest`:
```tsx
export const defineManifest = (manifest: ManifestV3Export): ManifestV3Export =>
  manifest
```

ok, that makes sense

FliPPeDround:
> hmm interesting. `defineManifest` doesnt do anything important?
> 
> update: yes, the whole source code for `defineManifest`:
> 
> ```tsx
> export const defineManifest = (manifest: ManifestV3Export): ManifestV3Export =>
>   manifest
> ```
> 
> ok, that makes sense

I might want to add this to the v2.0 release too.

jaskp:
@FliPPeDround This should be reverted. Now it's a type error to wrap the path in a function like `normalizePath` or getting it from env. Runtime validation during dev is early enough.

```typescript
import {defineManifest} from '@crxjs/vite-plugin'
import {normalizePath} from 'vite'

export default defineManifest(() => ({
  content_scripts: [
    {
      js: [normalizePath(process.env.CONTENT_SCRIPT_PATH)],
    },
  ],
}));

```

PR #1031, opened 2025-06-16
fix(playgrounds): rename playground packages and update ignore list

FliPPeDround:


FliPPeDround:
fix https://github.com/crxjs/chrome-extension-tools/pull/1030

PR #1045, opened 2025-07-14
fix: usage of newer picocolor api when using old dependency #1038

### LINKED ISSUE #1038: Calling functions from `picocolors@^1.1.0` but only requiring `^1.0.0`
### Build tool

Vite

- [ ] In the browser
- [x] In the terminal

### Describe the bug

The most recent updates to CRXJS use the new bright colour options from `picocolors`, but the [version specified](https://github.com/crxjs/chrome-extension-tools/blob/2a629cfe962c9a1deb2626193151498ad782bed6/packages/vite-plugin/package.json#L79) in `package.json` (`^1.0.0`) doesn't have these new bright colour variants. 

This causes the Vite plugin to fail on startup.

Please up the version to `^1.1.0` to make this go away 😄 

### Reproduction

N/A - private repo

### Logs

```shell

```

### System Info

```shell
System:
    OS: macOS 15.5
    CPU: (8) arm64 Apple M2
    Memory: 175.16 MB / 16.00 GB
    Shell: 5.9 - /bin/zsh
  npmPackages:
    @types/node: ^20.4.1 => 20.11.19 
    eslint: ^8.38.0 => 8.54.0 
    eslint-config-custom: 0.0.1 => 0.0.1 
    prettier: ^2.7.1 => 2.8.8 
    prettier-plugin-tailwindcss: ^0.1.13 => 0.1.13 
    syncpack: ^9.8.6 => 9.8.6 
    tsx: ^3.11.0 => 3.14.0 
    turbo: ^1.13.1 => 1.13.1 
    typescript: ^5.8.3 => 5.8.3
```

### Severity

blocking an upgrade

Toumash:
closes #1038

### DIFF
--- a/packages/vite-plugin/package.json
+++ b/packages/vite-plugin/package.json
@@ -76,7 +76,7 @@
     "jsesc": "^3.0.2",
     "magic-string": "^0.30.12",
     "pathe": "^2.0.1",
-    "picocolors": "^1.0.0",
+    "picocolors": "^1.1.1",
     "react-refresh": "^0.13.0",
     "rollup": "2.79.2",
     "rxjs": "7.5.7"
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -269,7 +269,7 @@ importers:
         specifier: ^2.0.1
         version: 2.0.3
       picocolors:
-        specifier: ^1.0.0
+        specifier: ^1.1.1
         version: 1.1.1
       react-refresh:
         specifier: ^0.13.0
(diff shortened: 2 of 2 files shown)


FliPPeDround:
@Toumash Hello, for such bug fixes, I think we can release a minor version. What do you think?

Toumash:
You mean to release this thing https://github.com/crxjs/chrome-extension-tools/pull/1044 ? Yeah

FliPPeDround:
yes, I think so

FliPPeDround:
I believe we can release minor updates for any issue affecting users or even for small optimizations. This allows users to access the latest improvements and support more promptly. Plus, frequent small updates shouldn’t disrupt their installation experience. What are your thoughts on this approach?

Toumash:
Yes, i was only unsure if you were referring to a minor bump vs patch bump or not.
Let's release as frequently as possible

FliPPeDround:
> Yes, i was only unsure if you were referring to a minor bump vs patch bump or not. Let's release as frequently as possible

like this `v2.0.2 -> v2.0.3`

Issue #1049, opened 2025-07-25
Non-inlined source maps should be included in web_accessible_resources


ligfx:
### Build tool

Vite

### Where do you see the problem?

- [x] In the browser
- [ ] In the terminal

### Describe the bug

Otherwise, they get blocked by `net::ERR_BLOCKED_BY_CLIENT`

### Reproduction

`vite.config.ts`:
```typescript
export default defineConfig({
    ...,
    build: {
        sourcemap: true,
    },
});

```

### Logs

```shell

```

### System Info

```shell
System:
  OS: macOS 15.5
  CPU: (14) arm64 Apple M4 Max
  Memory: 1.51 GB / 36.00 GB
  Shell: 5.9 - /bin/zsh
Binaries:
  Node: 24.4.0 - /opt/homebrew/bin/node
  Yarn: 1.22.22 - /opt/homebrew/bin/yarn
  npm: 11.4.2 - /opt/homebrew/bin/npm
Browsers:
  Chrome: 138.0.7204.169
  Safari: 18.5
npmPackages:
  @crxjs/vite-plugin: ^2.0.3 => 2.0.3 
  vite: ^7.0.5 => 7.0.6
```

### Severity

annoyance

w-geissinger:
I believe using inline sourcemaps should resolve your issue!


```ts
export default defineConfig({
    ...,
    build: {
        sourcemap: 'inline',
    },
});
```

It would be nice to have the standard sourcemaps working, but I've been operating on inline for a while with no major issues. Every once in a while, things can get a little funky in a monorepo setting.

ligfx:
@w-geissinger Right, this is an issue with separate sourcemaps, the inline sourcemaps work fine. In my case I’m trying to use separate sourcemaps.

PR #1050, opened 2025-07-28
feat: include separate sourcemaps in web_accessible_resources

ligfx:
Solve for #1049.

This fixes issues where separate sourcemaps get blocked with `net::ERR_BLOCKED_BY_CLIENT`


ligfx:
Not sure why the unit test is failing 🤔 Do sourcemaps differ between Unix and Windows?

@Toumash I added another commit which stops the test diff output from being truncated, could you re-run the workflow when you get a chance so I can see what the diff is please?

Toumash:
Nice, you did a fix for the truncated output - i couldnt make it work after may tries : D

It is about the newlines - strange because the isTextFile should remove them

Toumash:
Ah it's for the serve command. For build hmm
https://github.com/crxjs/chrome-extension-tools/blob/4980ef329ac43d68f209782560715883a1987547/packages/vite-plugin/tests/testOutput.ts#L92-L93

Toumash:
Try adding this thing:
[crxjs-map-snapshot-newlines.patch](https://github.com/user-attachments/files/21504234/crxjs-map-snapshot-newlines.patch)

Toumash:
That is strange - the test actually passes on my windows machine.
Probably because of this config - github runner has some strange default settings - different than a default git installation on windows causing the document to be converted into crlf on git clone instead of lf.
Lets trim the newlines using the patch above

```
PS C:\Users\toumash\Documents\test\chrome-extension-tools\packages\vite-plugin> git config get core.autocrlf
false
```

image

ligfx:
Added the patch. On my machine, it makes tests pass after manually converting the newlines to Windows-style CRLF. Hopefully it works on the runner too.

ligfx:
Needed to keep the backslash-descaping before the path replacement

Issue #1073, opened 2025-10-21
paths and types


Toumash:
@FliPPeDround This should be reverted. Now it's a type error to wrap the path in a function like `normalizePath` or getting it from env. Runtime validation during dev is early enough.

```typescript
import {defineManifest} from '@crxjs/vite-plugin'
import {normalizePath} from 'vite'

export default defineManifest(() => ({
  content_scripts: [
    {
      js: [normalizePath(process.env.CONTENT_SCRIPT_PATH)],
    },
  ],
}));

```

_Originally posted by @jaskp in https://github.com/crxjs/chrome-extension-tools/issues/1015#issuecomment-3425654591_

FliPPeDround:
When dealing with dynamic paths, if you need to ensure type correctness, it should be handled through TypeScript type annotations (e.g., `as ManifestFilePath`). However, this may not be crucial, and I might consider relaxing this restriction.

FliPPeDround:
Regarding types in general, they are currently maintained by us and may sometimes lag behind. I'm considering migrating to `@types/webextension-polyfill` in the future to obtain more comprehensive and up-to-date type definitions.

link: #903

jaskp:
@FliPPeDround Could we restrict the check to literal strings?

```ts
import type {IsStringLiteral} from 'type-fest'

type FilePathFields<T extends string> = {
  // ...
  options_page?: IsStringLiteral<T> extends true ? ManifestFilePath<T> : string
  // ... etc.
}
```

FliPPeDround:
> [@FliPPeDround](https://github.com/FliPPeDround) Could we restrict the check to literal strings?
> 
> import type {IsStringLiteral} from 'type-fest'
> 
> type FilePathFields<T extends string> = {
>   // ...
>   options_page?: IsStringLiteral<T> extends true ? ManifestFilePath<T> : string
>   // ... etc.
> }

I think so. Would you like to provide a PR?

PR #1074, opened 2025-10-22
feat(vite-plugin): limit path string checking to literals

jaskp:
https://github.com/crxjs/chrome-extension-tools/issues/1073

### DIFF
--- a/packages/vite-plugin/package.json
+++ b/packages/vite-plugin/package.json
@@ -117,6 +117,7 @@
     "rollup-plugin-dts": "^4.2.0",
     "rollup-plugin-esbuild": "4.10.3",
     "svelte": "^3.48.0",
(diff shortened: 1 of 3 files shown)


FliPPeDround:
thanks ;)

jaskp:
@FliPPeDround I wrongly assumed the build process and rollup-plugin-dts would inline the type import from type-fest, but it didn't, so I think we need to copy&paste the `IsStringLiteral` type from type-fest because it's just a devDependency, unless you know of a better way

FliPPeDround:
I think this usage can be placed in `dependencies`, since as a type method it participates in the runtime execution of the type.
Of course, if it's simple enough, duplicating a copy could also be a good alternative.

PR #1082, opened 2025-12-15
Update package manager from pnpm to bun

akemmanuel:
This PR updates the package manager from pnpm to bun, which offers several benefits:

🚀 **Performance Improvements:**
- Up to 2x faster dependency installation
- Significantly faster package resolution
- Reduced disk space usage

🛠️ **Developer Experience:**
- Single binary installation (no separate CLI)
- Built-in test runner and bundler
- Better TypeScript support out of the box

📦 **Migration Benefits:**
- Compatible with existing package.json files
- Same command structure (bun install, bun run build, etc.)
- Zero configuration needed

The changes replace all pnpm references with bun equivalents in the README.md file while maintaining the same functionality. This migration will improve development speed and reduce setup complexity for new contributors.

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,12 +41,12 @@ npm create crxjs@latest
 ## 💻 Development
 
 - Clone this repository
-- Install [pnpm](https://pnpm.io)
-- Install dependencies using `pnpm install`
-- Build the `vite-plugin` project using `pnpm build:vite-plugin`
-- Playgrounds project is located at `playgrounds/**`, using `pnpm play` can run the playgrounds
+- Install [bun](https://bun.sh)
+- Install dependencies using `bun install`
+- Build the `vite-plugin` project using `bun build:vite-plugin`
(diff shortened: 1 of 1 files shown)


Toumash:
@akemmanuel we wont have to update a thing? Just the readme?

akemmanuel:
You may need to make some changes to the package.json file and fix some errors that occur, but once everything has been migrated, it's much faster.

akemmanuel:
I did'nt check everything if it works

Toumash:
@akemmanuel but in general it works right? `pnpm run build` in `packages/vite-plugin` works just fine?

Toumash:
if so then lets update the PR with the workflows update to run bun

Toumash:
Much more changes needed for the bun to be supported in the project https://github.com/crxjs/chrome-extension-tools/pull/1095

PR #1091, opened 2026-01-10
fix: replace cheerio with node-html-parser and add vite peerDependency

### LINKED ISSUE #1086: npm deprecation warning - `whatwg-encoding`
### Build tool

Vite

- [ ] In the browser
- [x] In the terminal

### Describe the bug

npm warn deprecated whatwg-encoding@3.1.1: Use @exodus/bytes instead for a more spec-conformant and faster implementation    

### Reproduction

run npm install with package.json:
```
"devDependencies": {
    "@crxjs/vite-plugin": "^2.3.0",
}
```

### Logs

```shell

```

### System Info

```shell
npm warn deprecated whatwg-encoding@3.1.1: Use @exodus/bytes instead for a more spec-conformant and faster implementation
```

### Severity

annoyance

Toumash:
## Summary

- Replaces `cheerio` with `node-html-parser` to fix the whatwg-encoding deprecation warning (#1086)
- Adds `vite` as a peerDependency (`^3.0.0 || ^4.0.0 || ^5.0.0 || ^6.0.0 || ^7.0.0`)
- Adds `injectWorkspacePackages: true` to pnpm-workspace.yaml for proper peer resolution
- Updates vite7 test to use injected dependency

## Related

- Fixes #1086
- Alternative approach to #1090

### 1. Replaced cheerio with node-html-parser
The `cheerio` package depends on `whatwg-encoding` which is deprecated. Switched to `node-html-parser` which is lighter and doesn't have this dependency.

### 2. Added vite peerDependency
Without declaring `vite` as a peerDependency, pnpm couldn't create separate instances of the plugin for different vite versions. This caused the vite7 test to incorrectly use vite@3 instead of vite@7.

### 3. Added injectWorkspacePackages
This pnpm workspace setting enables proper peer dependency resolution, allowing separate virtual store entries:
- `@crxjs+vite-plugin@file+packages+vite-plugin_vite@3.2.11_...` (for vite-plugin's own tests)
- `@crxjs+vite-plugin@file+packages+vite-plugin_vite@7.3.1_...` (for vite7-manifest-test)

## Comparison with #1090

| Aspect | PR #1090 | This PR |
|--------|----------|---------|
| peerDependencies | Not declared | Explicitly declares vite ^3-^7 |
| Lockfile | Required manual editing | Auto-generates correctly |
| Workspace config | None | `injectWorkspacePackages: true` |
| Robustness | Fragile, breaks on fresh install | Robust, works automatically |

## Testing

All tests pass:
- ✅ 3 vite7 manifest tests
- ✅ 66 vite-plugin tests
- ✅ rollup-plugin tests


Toumash:
@FliPPeDround im not sure what peerDependencies change will do. Please help me out here

FliPPeDround:
Yes, exactly. This setup leverages `peerDependencies` to specify that a compatible Vite (version 3-7) must be present in the user's environment. If it's missing or out of range, running `npm install` will produce an error to notify them.

Toumash:
@FliPPeDround so you're saying it is alright, and if I for instance as a crxjs user, use vite 4, then it will just stay as it is meaning no additional user action is needed?.
That's cool. The thing is that it's even broken on the main branch because I managed to change pnpm-lock but not the package.json that creates the lock. So any time someone right now will clone the repo and pnpm install, the v7 tests won't work right now. The thing is: with the pnpm lock change on main the Vite 7 tests are working, but without it they don't.

FliPPeDround:
Yes, that's exactly right.

PR #1042, opened 2025-07-07
we need to initialize plugins as empty array

salmin89:
else you get this error:
```
error during build:
Build failed with 1 error:

[plugin crx:manifest-post] /@crx/manifest
TypeError: plugins is not iterable
```

PR #1092, opened 2026-01-11
fix: use configResolved to get plugins for rolldown-vite (Vite 7) compatibility

### LINKED ISSUE #1067: Support/help debug vite-rolldown beta
### Describe the problem

There are currently issues when trying to use with `vite-rolldown`. `vite-rolldown` is the next major version of `vite` and almost an order of magnitude faster building for many projects.

It would be good to make sure the tool is compatible - and potentially to identify any existing bugs with `vite-rolldown` (or more likely `rolldown`) so they can get fixed upstream.

### Describe the proposed solution

Add tests for `vite-rolldown`. Investigate and track issues as they get fixed.

I can likely help out here if there is interest - this issue is partially to see whether there is interest from maintainers.

### Alternatives considered

Using the standard `vite` is required today, so makes things slower and more likely issues will affect `crxjs` when `vite-rolldown` replaces `vite`.

### Importance

would make my life easier

Toumash:
## Summary

Fixes the "TypeError: plugins is not iterable" error when using rolldown-vite (Vite 7).

- In rolldown-vite (Vite 7's default bundler), the `buildStart` hook doesn't receive `options.plugins`
- This fix uses `configResolved` hook to get plugins, which works in both Vite 7 and older versions
- Keeps `buildStart` as fallback for older Vite versions
- Initializes plugins as empty array to prevent iteration errors

## Changes

1. **`packages/vite-plugin/src/node/plugin-manifest.ts`**: 
   - Initialize `plugins` as empty array instead of `undefined`
   - Add `configResolved` hook to grab plugins from resolved config

2. **New tests**:
   - `tests/vite7-plugins-test`: Tests with Vite 7 to verify `configResolved` fix works
   - `tests/vite3-plugins-test`: Tests backward compatibility with Vite 3

## Related

Inspired by PR #1042 - this implements a cleaner version of that fix without `@ts-ignore` comments.

Closes #1067

### DIFF
--- a/.github/workflows/vite-plugin.yml
+++ b/.github/workflows/vite-plugin.yml
@@ -7,14 +7,20 @@ on:
     branches: [main]
     paths:
       - "packages/vite-plugin/**"
+      - "tests/vite3-plugins-test/**"
       - "tests/vite7-manifest-test/**"
+      - "tests/vite7-plugins-test/**"
+      - "tests/vite7-e2e-test/**"
       - ".github/workflows/vite-plugin.yml"
       - "!**.md"
   pull_request:
     branches: [main]
     paths:
       - "packages/vite-plugin/**"
+      - "tests/vite3-plugins-test/**"
       - "tests/vite7-manifest-test/**"
+      - "tests/vite7-plugins-test/**"
+      - "tests/vite7-e2e-test/**"
       - ".github/workflows/vite-plugin.yml"
       - "!**.md"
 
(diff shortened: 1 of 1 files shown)


Toumash:
@salmin89 @jaskp @FliPPeDround

jaskp:
> In rolldown-vite (Vite 7's default bundler),

rolldown-vite is just a temporary fork of vite 7, now we should test against `vite@8.0.0-beta.7` instead, which actually does use rolldown by default

Toumash:
@jaskp okay, I've added the updated pull request. 
https://github.com/crxjs/chrome-extension-tools/pull/1096

Now I would merge this one #1092
Then merge the consolidation one #1093
And then merge the #1096 (adds the vite-8 beta tests you've mentioned)

PR #1095, opened 2026-01-11
chore: migrate from pnpm to bun package manager

Toumash:
## Summary

This PR completes the migration from pnpm to bun as the package manager for the monorepo.

- Replaces pnpm with bun (v1.2.15) as the package manager
- Updates all GitHub workflows to use `oven-sh/setup-bun@v2`
- Updates package.json scripts to use bun syntax (`bun run --cwd` instead of `pnpm -C`)
- Adds `bunfig.toml` for workspace configuration
- Removes `pnpm-lock.yaml` and `pnpm-workspace.yaml`
- Adds `bun.lock` (text format for better git diffs)
- Updates README with bun instructions

## Benefits

- **Faster installation**: ~6s vs ~30s for dependency installation
- **Simpler setup**: Single binary, no separate CLI installation
- **Compatible**: Works with existing `workspace:*` protocol for monorepo packages

## Testing

- [x] `bun install` completes successfully
- [x] `bun run build:vite-plugin` builds the plugin
- [x] `bun run test:run:out` passes all tests (66/66)
- [x] `bun run lint` passes

## Related

Closes the incomplete migration attempt in #1082


Toumash:
it doesnt change a thing for us so lets just drop it for now it was only an experiment

PR #1107, opened 2026-01-18
add: browser_specific_settings.gecko properties

null22277:
add browser_specific_settings.gecko optional properties for interface ManifestV3. If we register add-on on addons.mozilla.org, must define browser_specific_settings.gecko.data_collection_permissions.

### DIFF
--- a/.changeset/curvy-rockets-relate.md
+++ b/.changeset/curvy-rockets-relate.md
@@ -1,0 +1,5 @@
+---
+"@crxjs/vite-plugin": minor
+---
+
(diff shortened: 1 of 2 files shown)


Toumash:
@null22277 hello, and thank you for your contribution!

salmin89:
isn't data_collection_permissions optional in the manifest for firefox? we didn't use this and crxjs in 2.4.0 is now complaining about types missmatch

Toumash:
ah, to be fixed then

PR #1109, opened 2026-01-24
fix: copy CSS files declared in manifest content_scripts to output

### LINKED ISSUE #849: CSS in content_script is ignored.
### Build tool

Vite

- [X] In the browser
- [ ] In the terminal

### Describe the bug

if we specify the 
```
    "content_scripts": [
        {
            "matches": ["<all_urls>"],
            "css": ["src/content/overlay/style.css"],
            "js": ["src/content/overlay/content.ts"]
        }
    ],
 ```
 in the mainfest, the css part should be correctly handled by the vite plugin. However, it's compeletely ignored. and path is simply copied with the actual css file missing from the `dist` folder.

### Reproduction

add the "css" section the in the content_scripts and the issue can be reproduced.

### Logs

_No response_

### System Info

```shell
System:
    OS: Windows 11 10.0.22631
    CPU: (12) x64 12th Gen Intel(R) Core(TM) i5-1235U
    Memory: 4.97 GB / 15.83 GB
  Binaries:
    Node: 21.5.0 - C:\Program Files\nodejs\node.EXE
    npm: 10.2.4 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Edge: Chromium (120.0.2210.91)
    Internet Explorer: 11.0.22621.1
  npmPackages:
    @crxjs/vite-plugin: ^2.0.0-beta.21 => 2.0.0-beta.21
    vite: ^5.0.8 => 5.0.10
```

### Severity

annoyance

Toumash:
## Summary

Fixes #849 - CSS files specified directly in the manifest's `content_scripts.css` array were being ignored.

## Problem

When users declared CSS files in their manifest like:
```json
"content_scripts": [
    {
        "matches": ["<all_urls>"],
        "css": ["src/content/style.css"],
        "js": ["src/content/content.ts"]
    }
]
```

The CSS path was preserved in the output manifest, but the actual CSS file was **not copied** to the dist folder, causing the extension to fail to load the styles.

## Root Cause

In `packages/vite-plugin/src/node/plugin-manifest.ts`, the `assetTypes` array specifies which manifest files should be copied to the output. While `contentStyles` were extracted from the manifest in `files.ts`, they were never included in `assetTypes` and therefore never copied.

## Solution

Added `'contentStyles'` to the `assetTypes` array so CSS files declared in the manifest are properly copied to the output bundle.

## Testing

- Added output test (`vite-content-script-manifest-css`) that verifies CSS files are included in the build output
- Added e2e test (`mv3-content-script-manifest-css`) that launches a real Chrome browser and verifies the CSS is actually applied to the page
- All 51 output tests and 41 e2e tests pass

image

### DIFF
--- a/packages/vite-plugin/src/node/plugin-contentScripts.ts
+++ b/packages/vite-plugin/src/node/plugin-contentScripts.ts
@@ -15,7 +15,7 @@ import { basename } from './path'
 import { RxMap } from './RxMap'
 import { CrxPluginFn } from './types'
 import { contentHmrPortId, preambleId, viteClientId } from './virtualFileIds'
-import colors from 'picocolors';
+import colors from 'picocolors'
 
 /**
  * Emits content scripts and loaders.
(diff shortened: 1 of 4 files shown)


FliPPeDround:
I suddenly thought of a solution, though it's a bit hacky. Perhaps we can inject the CSS file paths from content_scripts directly into the JS file as import statements. This way, we can both ensure the files aren't lost and support hot reloading. What do you think? Is this feasible?

Toumash:
> I suddenly thought of a solution, though it's a bit hacky. Perhaps we can inject the CSS file paths from content_scripts directly into the JS file as import statements. This way, we can both ensure the files aren't lost and support hot reloading. What do you think? Is this feasible?

sure, this hack is possble, and this also means that the full solution is possible, but let's make it outside of this pull request

FliPPeDround:
Sure, I agree.

Issue #1120, opened 2026-02-18
Paths in tsconfig vs vite.config are wrong in vanilla exampel


salmin89:
### Build tool

Vite

### Where do you see the problem?

- [ ] In the browser
- [x] In the terminal

### Describe the bug

I used the getting started template from: https://crxjs.dev/guide/installation/create-crxjs/
and noticed that tsconfig has:
```
 "paths": {
      "~/*": ["src/*"]
    },
```

while vite.config.ts uses:
```
resolve: {
    alias: {
      '@': `${path.resolve(__dirname, 'src')}`,
    },
  },
```

### Reproduction

https://crxjs.dev/guide/installation/create-crxjs/

### Logs

```shell

```

### System Info

```shell
https://crxjs.dev/guide/installation/create-crxjs/
```

### Severity

blocking an upgrade

Toumash:
@FliPPeDround how we can publish the create-crxjs npm package? https://github.com/crxjs/create-crxjs/actions
It seems like it does not support the changesets workflow

FliPPeDround:
> [@FliPPeDround](https://github.com/FliPPeDround) how we can publish the create-crxjs npm package? [crxjs/create-crxjs/actions](https://github.com/crxjs/create-crxjs/actions) It seems like it does not support the changesets workflow

Certainly, let me address this issue. Due to the adjustments in npm publishing and the potential increase in packages from the crxjs organization, I plan to create a [crxjs/workflows](https://github.com/crxjs/workflows) project to handle such matters. However, since it is currently the Spring Festival holiday in China, I may not be able to complete this project immediately, but it won't be too long.

Toumash:
hm i dont think we need a external action. The create-crxjs just needs a tag push update from what i can see but i somehow cannot trigger it

PR #1121, opened 2026-02-18
corrected the tsconfig vanilla example paths

salmin89:


salmin89:
solves:
https://github.com/crxjs/chrome-extension-tools/issues/1120

PR #1147, opened 2026-03-30
feat(vite-plugin): add liveReload option to disable automatic extension reload and HMR

Toumash:
## Summary

Adds a `liveReload` option to `CrxOptions` that allows users to disable automatic extension reload and HMR during development. When `liveReload: false`, file changes still update the `outDir` on disk, but the extension does not automatically call `chrome.runtime.reload()`, push HMR updates to content scripts, or trigger `location.reload()` on host pages. Users can then manually reload the extension in Chrome when ready.

Closes https://github.com/crxjs/chrome-extension-tools/discussions/1146

## Usage

```ts
// vite.config.ts
import { crx } from '@crxjs/vite-plugin'

export default defineConfig({
  plugins: [
    crx({
      manifest,
      liveReload: false, // disable automatic reload/HMR
    }),
  ],
})
```

## Implementation

Three layers are gated by the option:

1. **Server side** (`plugin-hmr.ts`) — skips sending `crx:runtime-reload` and `crx:content-script-payload` WebSocket messages. The RxJS `crxHMRPayload$` subscription (which writes files to `outDir`) stays alive.
2. **Service worker client** (`hmr-client-worker.ts`) — `__LIVE_RELOAD__` build-time constant gates `chrome.runtime.reload()`, `notifyContentScripts()`, and reconnection-triggered reload.
3. **Content script client** (`hmr-content-port.ts`) — `__CRX_LIVE_RELOAD__` build-time constant gates `location.reload()`.

The option defaults to `true` (existing behavior preserved).

## Changes

- `src/node/types.ts` — Added `liveReload?: boolean` to `CrxOptions` with JSDoc
- `src/node/plugin-hmr.ts` — Gates WebSocket `send()` calls based on option
- `src/node/plugin-background.ts` — Injects `__LIVE_RELOAD__` into service worker client
- `src/node/plugin-contentScripts.ts` — Injects `__CRX_LIVE_RELOAD__` into content port client
- `src/client/es/hmr-client-worker.ts` — Guards reload calls with `__LIVE_RELOAD__`
- `src/client/es/hmr-content-port.ts` — Guards `location.reload()` with `__CRX_LIVE_RELOAD__`
- `.gitignore` — Broadened e2e test `src/` ignore pattern

## Tests

Two new e2e test fixtures:
- `mv3-vite-live-reload-default` — verifies default behavior: background change triggers runtime reload (options page closes, content page reloads)
- `mv3-vite-live-reload-disabled` — verifies `liveReload: false`: background change does NOT trigger reload (options page stays open, zero page reloads)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,7 +5,7 @@
 **/tests/artifacts
 .vscode
 **/chromium-data-dir-*
-**/tests/e2e/mv3-*-hmr/src
+**/tests/e2e/mv3-*/src
 **/__diff_output__
 TODO
 pnpm-global
--- a/packages/vite-plugin/src/client/es/hmr-client-worker.ts
+++ b/packages/vite-plugin/src/client/es/hmr-client-worker.ts
@@ -16,6 +16,7 @@ declare const __HMR_PORT__: string
 declare const __HMR_TIMEOUT__: number
 declare const __SERVER_PROTO__: string
 declare const __SERVER_PORT__: string
+declare const __LIVE_RELOAD__: boolean
 
 /* -------- REDIRECT FETCH TO THE DEV SERVER ------- */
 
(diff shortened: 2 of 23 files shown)


salmin89:
crxjs creates a -loader.js for hmr support. But in the case you have a content script that does not import/export, this is only true during dev. In build the loader is not necessary.

Would this option have this in mind? If we disable hmr, content script without imports should not have a loader?

Toumash:
In this PR we dont change the loader setup just not sending the HMR events.
It is only about the dev mode

FliPPeDround (approved):

PR #1149, opened 2026-04-04
fix: Vite 8 (Rolldown) compatibility for content scripts

sheldhur:
## Summary

Two fixes for Vite 8 (Rolldown) compatibility when building Chrome extensions with content scripts:

1. **Duplicate refIds**: `this.emitFile({ type: "chunk", name: basename(file) })` returns the same `refId` for different files with the same basename (e.g. `seller/content.tsx`, `cmp/content.tsx`, `marketplace/content.tsx`). Only the first script gets its `fileName` resolved; the rest fail with `Content script fileName is undefined`.

2. **Wrong MAIN world loader path**: The loader uses `./${fileName.split("/").at(-1)}` which loses the subdirectory, producing a broken path.

### DIFF
--- a/packages/vite-plugin/src/node/plugin-contentScripts.ts
+++ b/packages/vite-plugin/src/node/plugin-contentScripts.ts
@@ -204,7 +204,7 @@ export const pluginContentScripts: CrxPluginFn = () => {
                   }),
                   source: worldMainIds.has(script.id)
                     ? createProMainLoader({
-                        fileName: `./${fileName.split('/').at(-1)}`,
+                        fileName: `../${fileName}`,
                       })
                     : createProLoader({ fileName }),
                 })
--- a/packages/vite-plugin/src/node/plugin-manifest.ts
+++ b/packages/vite-plugin/src/node/plugin-manifest.ts
@@ -260,7 +260,7 @@ export const pluginManifest: CrxPluginFn = () => {
                 const refId = this.emitFile({
                   type: 'chunk',
                   id,
-                  name: basename(file),
+                  name: file.replace(/[\\/]/g, '-'),
                 })
                 contentScripts.set(
                   file,


sheldhur:
@Toumash can you look at this? I can update the failed tests if you're ready to merge it

silas:
I'm not sure if it's the same issue or not (I've haven't dug into it yet), but I also had an issue where I had a content script and service_worker script both named `main.ts` (but in different directories), and the generated `service-worker-loader.js` script ended up importing the content module/file instead of the service_worker module/file.

I'm using Vite 8.

sheldhur:
@Toumash ping

packages/vite-plugin/src/node/plugin-contentScripts.ts
```diff
@@ -204,7 +204,7 @@ export const pluginContentScripts: CrxPluginFn = () => {
                   }),
                   source: worldMainIds.has(script.id)
                     ? createProMainLoader({
-                        fileName: `./${fileName.split('/').at(-1)}`,
+                        fileName: `../${fileName}`,
```
Toumash, line 207: i'm not sure about the subdirectory why would it need a different behavior than a line below                    

```ts
 : createProLoader({ fileName }),
```

sheldhur: It just works this way. If you want to see something like `createProMainLoader({ fileName })` I can change `createProMainLoader`

Toumash:
Thanks @sheldhur ! I checked this against current `main`.

Below regression test fails on current `main`, and passes with only your two changes from this PR applied.

The test covers the reported content-script case:
- duplicate content script basenames like `seller/content.ts`, `cmp/content.ts`, `marketplace/content.ts`
- a nested `MAIN` world content script loader resolving to the real emitted chunk

It does not prove a fully generic fix for every manifest-emitted chunk collision. Since this PR is scoped to content scripts, I think that is fine. If we want to harden all manifest chunk emission later, that can be a separate follow-up.

Could you add this compat test? After #1179, `--mode compat` runs in the Vite matrix, so CI should measure it against Vite 8 too.

```ts
import {
  existsSync,
  mkdirSync,
  mkdtempSync,
  readFileSync,
  rmSync,
  writeFileSync,
} from 'node:fs'
import { tmpdir } from 'node:os'
import { dirname, join, normalize } from 'pathe'
import { build } from 'vite'
import { afterEach, expect, test } from 'vitest'
import { crx } from 'src/.'

let rootDir: string | undefined

function writeFixture(fileName: string, source: string) {
  const filePath = join(rootDir!, fileName)
  mkdirSync(dirname(filePath), { recursive: true })
  writeFileSync(filePath, source)
}

function getImportPath(source: string) {
  return (
    source.match(/import\s+["']([^"']+)["']/)?.[1] ??
    source.match(/import\([\s\S]*?["']([^"']+)["'][\s\S]*?\)/)?.[1]
  )
}

function resolveImportPath(importer: string, importPath: string) {
  return normalize(join(dirname(importer), importPath))
}

afterEach(() => {
  if (rootDir) rmSync(rootDir, { recursive: true, force: true })
  rootDir = undefined
})

test('builds duplicate-basename content scripts and nested MAIN world loaders', async () => {
  rootDir = mkdtempSync(join(tmpdir(), 'crxjs-content-script-paths-'))
  const distDir = join(rootDir, 'dist')

  writeFixture(
    'src/shared.ts',
    `export function log(message: string) { console.log(message) }`,
  )
  writeFixture(
    'src/seller/content.ts',
    `import { log } from '../shared'; export function onExecute() { log('seller-content') }`,
  )
  writeFixture(
    'src/cmp/content.ts',
    `import { log } from '../shared'; export function onExecute() { log('cmp-content') }`,
  )
  writeFixture(
    'src/marketplace/content.ts',
    `import { log } from '../shared'; export function onExecute() { log('marketplace-content') }`,
  )
  writeFixture('src/worker/content.ts', `console.log('background-content')`)

  await build({
    root: rootDir,
    configFile: false,
    envFile: false,
    logLevel: 'error',
    build: { outDir: distDir, minify: false },
    plugins: [
      crx({
        manifest: {
          manifest_version: 3,
          name: 'Content Script Path Compatibility Test',
          version: '1.0.0',
          background: { service_worker: 'src/worker/content.ts' },
          content_scripts: [
            { js: ['src/seller/content.ts'], matches: ['https://seller.example/*'] },
            { js: ['src/cmp/content.ts'], matches: ['https://cmp.example/*'] },
            {
              js: ['src/marketplace/content.ts'],
              matches: ['https://marketplace.example/*'],
              world: 'MAIN',
            },
          ],
        },
      }),
    ],
  })

  const manifest = JSON.parse(readFileSync(join(distDir, 'manifest.json'), 'utf8'))
  const contentScriptFiles = manifest.content_scripts.flatMap(({ js }) => js)

  expect(contentScriptFiles).toHaveLength(3)
  expect(new Set(contentScriptFiles).size).toBe(contentScriptFiles.length)

  const mainWorldLoader = manifest.content_scripts.find(
    ({ world }) => world === 'MAIN',
  ).js[0]
  const mainWorldLoaderSource = readFileSync(join(distDir, mainWorldLoader), 'utf8')
  const mainWorldChunkPath = resolveImportPath(
    mainWorldLoader,
    getImportPath(mainWorldLoaderSource)!,
  )

  expect(mainWorldChunkPath).toContain('marketplace')
  expect(existsSync(join(distDir, mainWorldChunkPath)), mainWorldChunkPath).toBe(true)
})
```

PR #1155, opened 2026-04-17
Allow full-reload HMR events with a path

flacoshinobi:
Removes the filter that blocked 'full-reload' events when a path was specfied. This ensures all full-reload signals are forwarded to the content scripts, improving compatibility with Vite 7+ and fixing cases where reloads were missed


Toumash:
hmm how about only implementing it only for vite7 then? I forgot what does it do and i remember fighting that filter in one of the e2e tests

flacoshinobi:
This update is also important for compatibility with Vite 8.
I found that starting from version 7.1.0-beta.0, Vite now uses `*` as the default value for the path during a full-reload.
As a result, the `undefined` value that the current logic expects is no longer passed in newer versions, which is why full-reload has stopped working in content scripts

You can see the specific change in the Vite source code here:
https://github.com/vitejs/vite/blob/v7.1.0-beta.0/packages/vite/src/node/server/hmr.ts#L712-L717

If there are reasons not to allow all paths, we should at least check for both `*` and `undefined` to ensure backward compatibility with older Vite versions while supporting Vite 7.1+ and Vite 8

Toumash:
now we have the full vite version compatibility checks in place - it is safe to update the branch and merge if checks pass

Toumash:
Hey @flacoshinobi, your diagnosis here looks correct: newer Vite versions can send `full-reload` with a path, including `*`, and CRXJS should not drop that event for content scripts.

Since this PR was last reviewed, `main` got a broader HMR/Vite compatibility update. The current setup now maps `full-reload` payloads and `shouldForwardCrxPayload()` only filters empty `prune` / `update` payloads, so `full-reload` with a path should be forwarded:

https://github.com/crxjs/chrome-extension-tools/blob/276be1dce7eeec141a89c7c68558f6400a2efe30/packages/vite-plugin/src/node/fileWriter-hmr.ts#L141

I checked the exact Vite wildcard case you mentioned with this unit test:

```ts
test('forwards Vite wildcard full reloads to content scripts', () => {
  const payload = mapVitePayloadForCrx({
    type: 'full-reload',
    path: '*',
  })

  expect(payload).toMatchObject({
    type: 'full-reload',
  })
  expect(shouldForwardCrxPayload(payload)).toBe(true)
})
```

Since the implementation is already covered on `main`, would you be open to updating this PR so we can merge it as a regression test for the Vite 7.1+ / Vite 8 wildcard `full-reload` case?

flacoshinobi:
Hi @Toumash, I've merged the latest main into the branch and added the regression test. Thanks!

PR #1186, opened 2026-06-09
feat: add MAIN world content script HMR

### LINKED ISSUE #1159: MAIN world HMR
### Describe the problem

The MAIN world support in #1080 doesn't include HMR support.

### Describe the proposed solution

We could use the existing HMR infrastructure, just with the catches that ports have to be qualified with the extension ID and the runtime hardened against malicious sites.

### Alternatives considered

I would be fine with full reloading; HMR would just be more convenient. Rolldown can provide appropriate speed for that.

### Importance

would make my life easier

Toumash:
## What changed

- Adds dev HMR support for manifest-declared `world: "MAIN"` content scripts by loading the Vite client from the MAIN-world loader.
- Routes MAIN-world HMR through a tokenized `chrome.runtime.connect` port to the extension service worker.
- Adds dev-only `externally_connectable.matches` for valid concrete MAIN-world content-script host matches, without broad `ids: ["*"]` access.
- Warns when Chrome cannot accept a content-script match pattern such as `<all_urls>` for `externally_connectable.matches`.
- Adds an e2e regression test proving MAIN-world content scripts update without page reload.

## Why

Manifest-declared MAIN-world content scripts previously fell back to reload behavior and logged that HMR was unsupported. The extension service worker already proxies Vite HMR, but MAIN-world scripts need an externally-connectable runtime port because they execute in the page context.

Fixes #1159.

## Validation

- `pnpm run lint:types`
- `pnpm run lint:eslint` (passes with the existing `plugin-manifest.ts` warning)
- `pnpm run build`
- `pnpm run test:run:out`
- `pnpm run test:run:compat`
- `pnpm run test:run:e2e`
- Targeted Vite 3 e2e and output tests for MAIN-world HMR and `externally_connectable` behavior

https://github.com/user-attachments/assets/5897d33a-1439-4d7c-9cb0-99b4f6a41209

### DIFF
--- a/packages/vite-plugin/src/client/es/hmr-content-port.ts
+++ b/packages/vite-plugin/src/client/es/hmr-content-port.ts
@@ -5,9 +5,31 @@ import type { HMRPayload } from 'vite'
 
 declare const __CRX_HMR_TIMEOUT__: number
 declare const __CRX_LIVE_RELOAD__: boolean
+declare const __CRX_HMR_TOKEN__: string
+
+const crxClientPortName = `@crx/client:${__CRX_HMR_TOKEN__}`
 
 function isCrxHMRPayload(x: HMRPayload): x is CrxHMRPayload {
   return x.type === 'custom' && x.event.startsWith('crx:')
+}
+
+function getRuntime() {
+  return typeof chrome === 'undefined' ? undefined : chrome.runtime
+}
+
+function getExtensionId() {
+  return new URL(import.meta.url).host
+}
+
+function hasOwnExtensionRuntime(
+  runtime: typeof chrome.runtime,
+  extensionId: string,
+) {
+  try {
+    return new URL(runtime.getURL('')).host === extensionId
+  } catch {
+    return false
+  }
 }
 
 export class HMRPort {
@@ -39,8 +61,16 @@ export class HMRPort {
   }
 
   initPort = () => {
+    const runtime = getRuntime()
+    if (!runtime) throw new Error('[crx] chrome.runtime is not available')
+
+    const extensionId = getExtensionId()
+    const connectInfo = { name: crxClientPortName }
+
     this.port?.disconnect()
-    this.port = chrome.runtime.connect({ name: '@crx/client' })
+    this.port = hasOwnExtensionRuntime(runtime, extensionId)
+      ? runtime.connect(connectInfo)
+      : runtime.connect(extensionId, connectInfo)
     this.port.onDisconnect.addListener(this.handleDisconnect.bind(this))
     this.port.onMessage.addListener(this.handleMessage.bind(this))
     this.port.postMessage({ type: 'connected' })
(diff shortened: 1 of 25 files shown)


packages/vite-plugin/src/client/es/hmr-content-port.ts
```diff
@@ -39,8 +61,16 @@ export class HMRPort {
   }
 
   initPort = () => {
+    const runtime = getRuntime()
```
gkgoat1: Maybe save the runtime? some sites delete `chrome` off the `window`

packages/vite-plugin/src/client/es/hmr-content-port.ts
```diff
@@ -5,11 +5,33 @@ import type { HMRPayload } from 'vite'
...
+function getExtensionId() {
+  return new URL(import.meta.url).host
+}
+
+function hasOwnExtensionRuntime(
+  runtime: typeof chrome.runtime,
+  extensionId: string,
+) {
+  try {
+    return new URL(runtime.getURL('')).host === extensionId
```
gkgoat1, line 17: This could also be cached

packages/vite-plugin/src/client/es/hmr-content-port.ts
```diff
@@ -5,11 +5,33 @@ import type { HMRPayload } from 'vite'
...
 function isCrxHMRPayload(x: HMRPayload): x is CrxHMRPayload {
   return x.type === 'custom' && x.event.startsWith('crx:')
 }
 
+function getRuntime() {
+  return typeof chrome === 'undefined' ? undefined : chrome.runtime
+}
+
+function getExtensionId() {
+  return new URL(import.meta.url).host
```
gkgoat1: Also cachable

gkgoat1 (commented):
LGTM, but some sites do weird things with chrome extension and browser APIs that I am personally wary of not keeping in mind--I added some comments along that line

packages/vite-plugin/src/client/es/hmr-content-port.ts
Toumash: this last one doesnt need to be as it is called only once with the working runtime thats also saved

Toumash:
Done. Since MAIN world code shares globals with the host page, we now cache `chrome.runtime` and the own-runtime check at module initialization. I also added an e2e regression that replaces `window.chrome` and verifies HMR still connects through the original extension runtime.

@gkgoat1 see the update

salmin89:
I tested it out and it seems to work nice.

Do we also support dynamic content scripts? 
If not, do we want to?

salmin89 (approved):

Toumash:
Dynamic scripts support would be cool, but i couldn't figure out how to do it easily in this PR. 
I want to ship features incrementally so that we have some shipping momentum

PR #1190, opened 2026-06-10
Add native content script HMR proof of concept

Toumash:
## Summary

Adds an experimental `contentScripts.hmr: "native"` mode for module content scripts. In this mode, the dev content-script loader imports Vite's native HMR client and the content-script module directly from the Vite dev server instead of using CRXJS's transformed content-script module and custom HMR-port bridge.

This is a proof of concept for #993. It does not remove the existing file-writer path; the current behavior remains the default.

## Compatibility coverage

Native mode is enabled in these existing e2e fixtures as compatibility coverage:

- `mv3-vite-vanilla-content-script-hmr`
- `mv3-vite-vanilla-content-script-seq-hmr`
- `mv3-vite-react-content-script-hmr`
- `mv3-vite-svelte-content-script-hmr`
- `mv3-vite-vue-content-script-hmr`
- `mv3-content-script-manifest-css-hmr`

The dedicated native-mode fixture remains in place as a small direct proof that a content-script module can receive Vite HMR over WebSocket.

## Notes

Native mode still writes the extension shell needed for development, including `manifest.json`, loader files, public assets, HTML placeholders, and IIFE bundles where applicable. The reduction is specifically in the content-script HMR path: native mode no longer needs to write the transformed module content script or patched Vite client files for module content scripts.

The measurable value in this POC is architectural simplification rather than clear speed or memory wins. Small fixture timings and peak RSS were effectively similar between native and file-writer mode.

Some fixtures intentionally remain on file-writer mode because they either assert disk-written virtual CSS files or expose follow-up native-mode gaps:

- `mv3-vite-virtual-css-hmr`
- `mv3-vite-unocss-hmr`
- `mv3-vite-dynamic-content-script-hmr`
- `mv3-vite-react-content-script-seq-hmr`

## Validation

- `pnpm -C packages/vite-plugin run lint:types`
- `pnpm -C packages/vite-plugin exec vitest --run --mode unit src/node/plugin-extensionCors.test.ts`
- `pnpm -C packages/vite-plugin exec vitest --run --mode out tests/out/native-content-script-hmr/serve.test.ts`
- `pnpm -C packages/vite-plugin exec vitest --run --mode e2e tests/e2e/mv3-vite-native-content-script-hmr/vite-serve.test.ts`
- Sequential e2e run of the six existing native-enabled fixtures listed above

PR #1196, opened 2026-06-12
fix: HRM and service script loader causes modules to be loaded twice with vite

### LINKED ISSUE #869: HRM and service script loader causes modules to be loaded twice with vite
### Build tool

Rollup

- [X] In the browser
- [ ] In the terminal

### Describe the bug

Imagine 2 files:
1.  `entry.js` referenced and load from `service-script-loader.js` that imports `another.js`
2.  and `another.js` imported to `entry.js`. 

So this 2 files reference each other, that happens but not an issue since postpone any execution `runtime.onInstall`, and it all works on initial `vite dev`.

Problems starts when update one of the files. If you update `another.js` the `entry.js`, the plugin will reload  and the following chain of events will happed.
1. `service-script-loader.js` will import un-timestamped `entry.js` since loader is only emitted once at the launch of `vite`
2. `entry.js` will reference and import timestamped `another.js?t=xxx` since `entry.js` is dynamically generated.  
3. `another.js?t=xxx` will import timestamped `entry.js?t=xxx`

Browser thinks that `entry.js` and `entry.js?t=xxx` are two different modules and executes the code twice.

I'm not familiar with rollup and vite plugins enough to attempt to fix it myself, but if someone is wiling to give me advice on how to reemit `service-script-loader.js` on HRM event, I'm  wiling to try.

For now the workaround I use is to add static one line entry point that only `import './entry.js'`

### Reproduction

https://github.com/jurijsk/crxjs-hrm-loader-issue

### Logs

_No response_

### System Info

```shell
System:
    OS: Windows 10 10.0.19045
    CPU: (16) x64 12th Gen Intel(R) Core(TM) i7-1260P
    Memory: 16.37 GB / 31.71 GB
  Binaries:
    Node: 18.18.0 - C:\Program Files\nodejs\node.EXE
    npm: 9.8.1 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Edge: Chromium (121.0.2277.128), ChromiumDev (123.0.2400.1)
    Internet Explorer: 11.0.19041.3636
```

### Severity

annoyance

Toumash:
Fixes #869.

## Summary
- Update the dev background loader on background hot updates so the worker entry URL receives Vite's HMR timestamp.
- Add an MV3 serve-mode regression fixture for the circular background module graph that previously registered duplicate listeners.

## Root cause
During a dev reload, the service worker loader kept importing the unversioned background entry while Vite rewrote the circular dependency graph with a `?t=...` timestamp. The browser treated those as distinct module URLs, so background entry side effects could run twice in one worker instance.

## Validation
- `pnpm --dir packages/vite-plugin exec vitest --run --mode e2e tests/e2e/mv3-vite-background-module-cache/vite-serve.test.ts tests/e2e/mv3-vite-live-reload-default/vite-serve.test.ts tests/e2e/mv3-vite-live-reload-disabled/vite-serve.test.ts tests/e2e/mv3-vite-hmr-background-define/vite-serve.test.ts`
- `pnpm --dir packages/vite-plugin run lint:types`
- `pnpm --dir packages/vite-plugin run lint:eslint`

### DIFF
--- a/packages/vite-plugin/src/node/plugin-background.ts
+++ b/packages/vite-plugin/src/node/plugin-background.ts
@@ -1,12 +1,25 @@
 import workerHmrClient from 'client/es/hmr-client-worker.ts'
+import fs from 'fs-extra'
 import type { ResolvedConfig } from 'vite'
 import { defineClientValues } from './defineClientValues'
-import { getFileName } from './fileWriter-utilities'
+import { getFileName, getOutputPath } from './fileWriter-utilities'
 import { getCrxHmrToken } from './hmrToken'
+import { isImporter } from './isImporter'
 import { ChromeManifestBackground, FirefoxManifestBackground } from './manifest'
+import { join, normalize } from './path'
 import { getOptions } from './plugin-optionsProvider'
 import type { Browser, CrxPluginFn, ResolvedConfigWithHMRToken } from './types'
 import { workerClientId } from './virtualFileIds'
+
+function addTimestamp(url: string, timestamp?: number): string {
+  if (!timestamp) return url
+  const [pathname, query] = url.replace(/[?&]t=\d+/, '').split('?')
+  return `${pathname}?t=${timestamp}${query ? `&${query}` : ''}`
+}
+
+function escapeRegExp(text: string) {
+  return text.replace(/[.*+?^${}()|[\]\\]/g, '\\$&')
+}
 
 /**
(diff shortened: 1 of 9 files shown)


Toumash:
Checked against #1190: that PR is scoped to native content-script HMR and does not update the dev background loader. This PR still applies as-is; no changes are needed after #1190.

PR #1202, opened 2026-06-13
feat: support vite 8 plugin builds

Toumash:
## Summary

- build the Vite plugin against Vite 8 and current TypeScript/Node types
- keep the workflow model where the plugin is built first, then the test matrix installs the Vite version under test, including Vite 3
- replace Rollup public output/context types in plugin internals with local structural types that work across Vite/Rollup/Rolldown boundaries
- update Vite 8 output snapshots and normalize Rolldown hashes/runtime helper chunks in snapshot tests

## Notes

Vite 8 production builds run through Rolldown. Some fixtures now emit a `rolldown-runtime` helper chunk for shared CJS/ESM interop when the test runner forces a vendor chunk. The tests keep that file in the output list but skip snapshotting its generated body, same as vendor output.

## Validation

- `pnpm -C packages/vite-plugin run test:run:out`
- `pnpm -C packages/vite-plugin exec tsc --noEmit`
- `pnpm -C packages/vite-plugin run lint:eslint`
- `pnpm -C packages/vite-plugin run build`
- `pnpm -C packages/vite-plugin run test:run:compat`
- temporarily installed `vite@3.2.11` and ran `vitest --run --mode compat`


packages/vite-plugin/src/node/plugin-contentScripts_dynamic.ts
```diff
@@ -69,9 +69,13 @@ export const pluginDynamicContentScripts: CrxPluginFn = () => {
               dynamicScripts.push(script)
             }
           }
+          if (dynamicScripts.length === 0) return
```
salmin89, line 72: Possible issues here:

silent semantic change in cleanup

The old code unconditionally called contentScripts.clear() on every dynamic build pass. The new code does two things at once:
- Early-returns if no dynamic scripts in this build pass
- Only deletes entries with isDynamicScript

Combined effect: in watch mode, if a build previously had dynamic scripts and the next build has none, the stale dynamic entries never get cleared. This is a behavior change worth either fixing or commenting.

packages/vite-plugin/src/node/plugin-hmr.ts
```diff
@@ -71,7 +71,7 @@ export const pluginHMR: CrxPluginFn = () => {
       // server should ignore outdir
       configResolved(_config) {
         config = _config
-        const { watch = {} } = config.server
+        const watch = config.server.watch ?? {}
```
salmin89, line 74: Vite 8 lets users set server.watch: null to disable chokidar entirely. The new config.server.watch ?? {} overrides that and re-enables a watcher the user disabled. The old destructure-default suffered the same bug, but it's worth fixing now while you're in the file.

packages/vite-plugin/package.json
```diff
@@ -70,7 +70,7 @@
     "test:e2e": "vitest --mode e2e",
     "test:run": "run-s test:run:*",
     "test:run:compat": "vitest --run --mode compat",
-    "test:run:out": "vitest --run --mode out",
+    "test:run:out": "vitest --run --mode out --threads false",
```
salmin89, line 73: why is  threads false here?

salmin89:
NIT (ai generated, feel free to ignore)

### Type-safety concerns (the new structural types in types.ts)

5. CrxOutputChunk is missing viteMetadata — this is load-bearing for any plugin that reads chunk.viteMetadata.importedCss. If your downstream code (or third-party renderCrxManifest consumers) touches it, they now compile against an empty type.

6. CrxOutputChunk.modules: Record<string, unknown> — obliterates real RenderedModule typing; iterating .code/.renderedLength gets no checks. Either type properly or drop it.

7. CrxOutputChunk.facadeModuleId: string | null — Rolldown can return undefined in intermediate states. Widen to string | null | undefined.

8. as unknown as casts proliferate — plugin-manifest.ts:476-483, plugin-contentScripts.ts:272-274, plugin-contentScripts_iife.ts:121-126. These bypass the type system permanently. Once a downstream CrxPluginContext field gets added, the casts silently keep compiling.

### Test infrastructure smells

9. Rolldown runtime chunk is listed but body skipped (testOutput.ts:103) — defensible for snapshot stability, but there's no counterbalancing assertion that it's correctly wired into the entry chunks. If Rolldown changes the wiring and it breaks at runtime, this test passes.

10. Hash regex broadened to [a-zA-Z0-9_-]+ with a hardcoded {5} length (testOutput.ts:67) — if Rolldown ever changes id length, snapshot normalization silently breaks and you get flaky CI.

11. No unit tests for the new duplicateBasenames / getContentScriptChunkName collision logic in plugin-manifest.ts:74-96 — trivially testable, 30 lines, three branches.


### Smaller things

- fileWriter.ts:60-72 strips output/platform from rollupOptions via denylist before calling rollup(). Allowlist would be safer — Vite 8 may add more Rolldown-only fields.
- plugin-htmlInlineScripts.ts:60-71: as TransformIndexHtmlObject cast short-circuits the discriminator. 'handler' in p.transformIndexHtml would be cleaner.
- plugin-manifest.ts:524 has a stale comment "this appears to run after generateBundle since this is the last plugin" — if you don't know when it runs, find out.
- svelteTestUtils.ts:13: the new \bbadd_location\$?\( regex now handles both minified and un-minified output; worth a one-line comment.
- No changeset entry visible in the diff.

salmin89:
One thing worth confirming: the Vite-3 branches in plugin-manifest.ts:579-586 and plugin-webAccessibleResources.ts:244-252 were written against Vite 3's ResolvedConfig shape and are now type-checked against Vite 8's. 

Are the branch conditions still meaningfully detecting Vite 3, or have they degraded to always-true/always-false under the new types? 

If you've already checked this you can ignore this comment. 

----

If not you can add this, run it, just to confirm. No need to commit the log however :) 

```typescript
// packages/vite-plugin/src/node/plugin-manifest.ts 

if (
  viteMajorVersion < 4 &&
  filename.endsWith('.map') &&
  config.build.sourcemap === true
) {
  console.log('[crx-debug] vite-3 branch hit: plugin-manifest sourcemap skip', { filename, viteMajorVersion })
  return
}
```


```typescript
// packages/vite-plugin/src/node/plugin-webAccessibleResources.ts

if (
  sourcemapFileName in bundle ||
  (viteMajorVersion < 4 && config.build.sourcemap == true)
) {
  if (viteMajorVersion < 4 && !(sourcemapFileName in bundle)) {
    console.log('[crx-debug] vite-3 branch hit: webAccessibleResources sourcemap fallback', { sourcemapFileName, viteMajorVersion })
  }
  resourcesWithMaps.push(sourcemapFileName)
}
```

PR #1228, opened 2026-07-29
fix(vite-plugin): resolve @fs asset paths in dev mode for monorepo workspace packages

benhaooo:
## Problem

In a pnpm monorepo, when a workspace package (e.g. `packages/shared-assets`) imports image/SVG assets from its own source tree, Vite serves those files via `/@fs/<absolute-path>` URLs because they live outside the extension app's Vite root.

During `pnpm dev`, CRXJS calls `prepAsset` to read and write these files to the output directory. The file read used:

```ts
readFile(join(server.config.root, id))
```

`pathe.join` (like Node's `path.join`) does **not** treat a path starting with `/@fs/…` as absolute — it concatenates, producing:

```
<root>/@fs/<absolute-path-to-asset>   ← file does not exist → crash
```

### Reproduction

Structure:
- `apps/extension` — CRXJS + Vite extension app (Vite root)
- `packages/shared-assets` — source-only workspace package; its `index.tsx` imports `badge.svg` and `tile.png` from `./assets/`

`pnpm dev` crashes immediately after Vite starts. The failing path is rooted under the extension app:
```
apps/extension/@fs/…/packages/shared-assets/src/assets/badge.svg
```
instead of the real absolute path.

## Fix

Mirror the existing `resolveScriptInput` helper (used for IIFE bundling) — strip the `/@fs` prefix to recover the absolute filesystem path:

```ts
const filePath = id.startsWith('/@fs/')
  ? id.slice('/@fs/'.length - 1)
  : join(server.config.root, id)
```

`'/@fs/'.length - 1` = 4, so `slice(4)` on `/@fs/Users/…` yields `/Users/…` (preserves the leading `/` for an absolute POSIX path). All other id forms fall through to the original logic unchanged.

## Verification

Tested with a minimal pnpm monorepo reproducer: <https://github.com/benhaooo/repro-crxjs-monorepo-assets>

`pnpm dev` runs without crashing and correctly writes the asset files.

### DIFF
--- a/.changeset/fix-prepasset-at-fs-path.md
+++ b/.changeset/fix-prepasset-at-fs-path.md
@@ -1,0 +1,5 @@
+---
+"@crxjs/vite-plugin": patch
+---
+
+Fix dev server crash when a monorepo workspace package imports assets (images, SVGs, etc.) from its own source tree outside the Vite root. `prepAsset` was reading the file with `join(server.config.root, id)`, but `pathe.join` does not treat a `/@fs/…` Vite URL as an absolute path and produces an incorrect path like `<root>/@fs/<abs-path>`. Strip the `/@fs` prefix to recover the real absolute filesystem path before calling `readFile`.
--- a/packages/vite-plugin/src/node/fileWriter-rxjs.ts
+++ b/packages/vite-plugin/src/node/fileWriter-rxjs.ts
@@ -1,3 +1,4 @@
+import { fileURLToPath } from 'url'
 import * as lexer from 'es-module-lexer'
 import fsx from 'fs-extra'
 import { readFile } from 'fs/promises'
@@ -211,7 +212,7 @@ function prepAsset(
       mergeMap(async ({ server }) => {
         const target = getOutputPath(server, fileName)
         const filePath = id.startsWith('/@fs/')
-          ? id.slice('/@fs/'.length - 1)
+          ? fileURLToPath('file://' + id.slice('/@fs'.length))
           : join(server.config.root, id)
         return {
           target,
--- a/packages/vite-plugin/src/node/fileWriter-rxjs.ts
+++ b/packages/vite-plugin/src/node/fileWriter-rxjs.ts
@@ -210,9 +210,12 @@ function prepAsset(
     $.pipe(
       mergeMap(async ({ server }) => {
         const target = getOutputPath(server, fileName)
+        const filePath = id.startsWith('/@fs/')
+          ? id.slice('/@fs/'.length - 1)
+          : join(server.config.root, id)
         return {
           target,
-          source: source ?? (await readFile(join(server.config.root, id))),
+          source: source ?? (await readFile(filePath)),
           deps: [],
         }
       }),
--- a/packages/vite-plugin/tests/out/vite-workspace-assets-shared/badge.svg
+++ b/packages/vite-plugin/tests/out/vite-workspace-assets-shared/badge.svg
@@ -1,0 +1,3 @@
+<svg xmlns="http://www.w3.org/2000/svg" aria-label="workspace asset">
+  <circle cx="8" cy="8" r="8" />
+</svg>
--- a/packages/vite-plugin/tests/out/vite-workspace-assets-shared/index.ts
+++ b/packages/vite-plugin/tests/out/vite-workspace-assets-shared/index.ts
@@ -1,0 +1,3 @@
+import badgeUrl from './badge.svg'
+
+export { badgeUrl }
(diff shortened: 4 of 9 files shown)


Toumash:
interesting find, i wonder what type of test would be best here that produces that same case as monorepo usage.
Let's see if making the file outside the root reproduces it

Toumash:
yup, it reproduces. 
Thanks for contribution @benhaooo! 
Please add the test case and lets merge then 

[chrome-extension-tools-355d374-test(vite-plugin): cover workspace assets outside root.patch](https://github.com/user-attachments/files/30558209/chrome-extension-tools-355d374-test.vite-plugin.cover.workspace.assets.outside.root.patch)

Toumash (changes requested):
looks great, lets add the test case and we'll merge

benhaooo:
> yup, it reproduces. Thanks for contribution @benhaooo! Please add the test case and lets merge then
> 
> [chrome-extension-tools-355d374-test(vite-plugin): cover workspace assets outside root.patch](https://github.com/user-attachments/files/30558209/chrome-extension-tools-355d374-test.vite-plugin.cover.workspace.assets.outside.root.patch)

thanks for the patch! 
applied the test case and pushed

Toumash:
it looks like the windows pipeline is red because of the duplicated drive name somehow


`'D:\D:\a\chrome-extension-tools\chrome-extension-tools\packages\vite-plugin\tests\out\vite-workspace-assets-shared\badge.svg'`

https://github.com/crxjs/chrome-extension-tools/actions/runs/30640199662/job/91208391410?pr=1228

```
Error: ENOENT: no such file or directory, open 'D:\D:\a\chrome-extension-tools\chrome-extension-tools\packages\vite-plugin\tests\out\vite-workspace-assets-shared\badge.svg'
Vitest caught 1 unhandled error during the test run.
This might cause false positive tests. Resolve unhandled errors to make sure your tests are not affected.
 ❯ open node:internal/fs/promises:639:25
 ❯ Proxy.readFile node:internal/fs/promises:1252:14
 ❯ src/node/fileWriter-rxjs.ts:218:30
    216|         return {
    217|           target,
    218|           source: source ?? (await readFile(filePath)),
       |                              ^
    219|           deps: [],
    220|         }

⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯
Serialized Error: { errno: -4058, code: 'ENOENT', syscall: 'open', path: 'D:\D:\a\chrome-extension-tools\chrome-extension-tools\packages\vite-plugin\tests\out\vite-workspace-assets-shared\badge.svg' }
This error originated in "tests/out/vite-workspace-assets/serve.test.ts" test file. It doesn't mean the error was thrown inside the file itself, but while it was running.
The latest test that might've caused the error is "serves assets imported by a workspace package outside the Vite root". It might mean one of the following:
- The error was thrown, while Vitest was running this test.
- This was the last recorded test before the error was thrown, if error originated after test finished its execution.

```

benhaooo:
> it looks like the windows pipeline is red because of the duplicated drive name somehow
> 
> `'D:\D:\a\chrome-extension-tools\chrome-extension-tools\packages\vite-plugin\tests\out\vite-workspace-assets-shared\badge.svg'`
> 
> https://github.com/crxjs/chrome-extension-tools/actions/runs/30640199662/job/91208391410?pr=1228
> 
> ```
> Error: ENOENT: no such file or directory, open 'D:\D:\a\chrome-extension-tools\chrome-extension-tools\packages\vite-plugin\tests\out\vite-workspace-assets-shared\badge.svg'
> Vitest caught 1 unhandled error during the test run.
> This might cause false positive tests. Resolve unhandled errors to make sure your tests are not affected.
>  ❯ open node:internal/fs/promises:639:25
>  ❯ Proxy.readFile node:internal/fs/promises:1252:14
>  ❯ src/node/fileWriter-rxjs.ts:218:30
>     216|         return {
>     217|           target,
>     218|           source: source ?? (await readFile(filePath)),
>        |                              ^
>     219|           deps: [],
>     220|         }
> 
> ⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯
> Serialized Error: { errno: -4058, code: 'ENOENT', syscall: 'open', path: 'D:\D:\a\chrome-extension-tools\chrome-extension-tools\packages\vite-plugin\tests\out\vite-workspace-assets-shared\badge.svg' }
> This error originated in "tests/out/vite-workspace-assets/serve.test.ts" test file. It doesn't mean the error was thrown inside the file itself, but while it was running.
> The latest test that might've caused the error is "serves assets imported by a workspace package outside the Vite root". It might mean one of the following:
> - The error was thrown, while Vitest was running this test.
> - This was the last recorded test before the error was thrown, if error originated after test finished its execution.
> ```

used fileURLToPath to handle the path uniformly,  let's try again...

Toumash:
running

PR #1232, opened 2026-08-11
fix(vite-plugin): replace all live reload placeholders

### LINKED ISSUE #1230: Only the first `__LIVE_RELOAD__` is substituted, so the worker throws on reconnect after a dev server restart
## Summary

`__LIVE_RELOAD__` is used **twice** in the HMR worker client, but the build substitutes it with a single `String.prototype.replace`, which replaces only the first occurrence. The survivor sits in the socket `close` handler, so **every dev server restart ends in a `ReferenceError` instead of the runtime reload that would reconnect HMR** — and because the build itself keeps succeeding, edits appear to keep working while HMR is already dead.

## Where

- `packages/vite-plugin/src/client/es/hmr-client-worker.ts` uses it at [L149](https://github.com/crxjs/chrome-extension-tools/blob/main/packages/vite-plugin/src/client/es/hmr-client-worker.ts#L149) (`handleCrxHmrPayload`) and again at [L189](https://github.com/crxjs/chrome-extension-tools/blob/main/packages/vite-plugin/src/client/es/hmr-client-worker.ts#L189), inside `socket.addEventListener('close', ...)`.
- `packages/vite-plugin/src/node/plugin-background.ts` substitutes it at [L45](https://github.com/crxjs/chrome-extension-tools/blob/main/packages/vite-plugin/src/node/plugin-background.ts#L45):

```js
.replace('__LIVE_RELOAD__', JSON.stringify(liveReload))
```

Present in the published `2.7.1` bundle (`dist/index.mjs`) and still present on `main` as of 2026-08-02.

## Why the second one is the one that matters

The reconnect path is the only place that runs it:

```js
socket.addEventListener('close', async ({ wasClean }) => {
  if (wasClean) return
  await waitForSuccessfulPing()
  if (__LIVE_RELOAD__) {          // <- never substituted
    handleCrxHmrPayload({ type: 'custom', event: 'crx:runtime-reload' })
  }
})
```

The first occurrence is substituted and works, so ordinary HMR looks fine. The second is only reached after the socket closes uncleanly and the server comes back — i.e. exactly when the worker is supposed to call `chrome.runtime.reload()` and re-attach.

## Reproduce

1. `vite` dev server up, extension loaded, HMR working.
2. Kill the dev server and start it again (same port).
3. Service worker console:

```
Uncaught (in promise) ReferenceError: __LIVE_RELOAD__ is not defined
    at WebSocket.<anonymous> (chrome-extension://<id>/@crx/client-worker:113:3)
```

The worker never reloads the extension. Builds keep landing in the output folder, so nothing else reports a problem.

This is not theoretical: the last two entries our extension's own error log recorded before we migrated away were exactly this, from the daily development browser.

## Fix

```diff
-        .replace('__LIVE_RELOAD__', JSON.stringify(liveReload))
+        .replaceAll('__LIVE_RELOAD__', JSON.stringify(liveReload))
```

Worth checking the neighbouring substitutions in the same file for the same shape — the defect class is "a placeholder that occurs more than once, substituted with a single `replace`", and a leftover reaches the browser as an undefined identifier rather than as a build error.

I don't have a stake in this any more (we've moved off CRXJS for unrelated reasons), so I'm not opening a PR — but the diagnosis above is complete and the one-word fix is above. Happy to answer questions if anything is unclear.

shixi-li:
## Summary

- replace every `__LIVE_RELOAD__` placeholder in the generated background worker client
- add a hook-level regression test that verifies no live-reload placeholder reaches the served client
- add a patch changeset for `@crxjs/vite-plugin`

The worker client uses this placeholder in both the HMR payload path and the dev-server reconnect path. A single string replacement left the reconnect occurrence undefined at runtime. The global regular expression keeps compatibility with the project's Node 14 minimum while replacing both occurrences.

Fixes #1230.

## Validation

- `pnpm --dir packages/vite-plugin exec vitest --run src/node/plugin-background.test.ts`
- `pnpm --dir packages/vite-plugin run test:run:out` (113 passed, 1 skipped)
- `pnpm --dir packages/vite-plugin run lint`
- `pnpm --dir packages/vite-plugin run lint:attw`
- Prettier and `git diff --check`

### DIFF
--- a/.changeset/calm-workers-reconnect.md
+++ b/.changeset/calm-workers-reconnect.md
@@ -1,0 +1,6 @@
+---
+'@crxjs/vite-plugin': patch
+---
+
+Replace every live reload placeholder in the background worker client so it can
+reconnect after a dev server restart.
--- a/packages/vite-plugin/src/node/plugin-background.test.ts
+++ b/packages/vite-plugin/src/node/plugin-background.test.ts
@@ -1,0 +1,43 @@
+import type { PluginContext } from 'rollup'
+import type { ResolvedConfig } from 'vite'
+import { expect, test } from 'vitest'
+import { pluginBackground } from './plugin-background'
+import type { CrxPlugin } from './types'
+import { workerClientId } from './virtualFileIds'
+
+test('replaces every live reload placeholder in the worker client', async () => {
+  const plugins = pluginBackground()
+  if (!Array.isArray(plugins)) {
+    throw new Error('Expected background plugins')
+  }
+
+  const clientPlugin = plugins.find(
+    (plugin): plugin is CrxPlugin => plugin.name === 'crx:background-client',
+  )
+  if (!clientPlugin) throw new Error('Unable to find background client plugin')
+
+  const loaderPlugin = plugins.find(
+    (plugin): plugin is CrxPlugin =>
+      plugin.name === 'crx:background-loader-file',
+  )
+  if (!loaderPlugin) throw new Error('Unable to find background loader plugin')
+
+  const configResolved = loaderPlugin.configResolved
+  if (typeof configResolved !== 'function') {
+    throw new Error('Unable to find configResolved hook')
+  }
+  await configResolved({
+    base: '/',
+    define: {},
+    mode: 'development',
+    server: { hmr: {}, https: false, port: 5173 },
+    webSocketToken: 'test-token',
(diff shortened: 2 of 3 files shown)


Toumash:
@shixi-li thanks for the contribution! Lets see if the tests will pass

PR #1238, opened 2026-08-29
fix: use Vite 8 WebSocket config

### LINKED ISSUE #1231: Deprecation warning on startup: migrate from server.hmr to server.ws for Vite 8
### Build tool

Vite

- [ ] In the browser
- [x] In the terminal

### Describe the bug

Every time the project starts, `@crxjs/vite-plugin` triggers a Vite 8 deprecation warning:

> `server.hmr.protocol/host/port/path/clientPort/timeout/server` is deprecated. Use `server.ws.*` instead. Note that this option may be set by a plugin. Set VITE_DEPRECATION_TRACE=1 to see where it is called.

The root cause is that `@crxjs/vite-plugin` still accesses `config.server.hmr` in its `defineClientValues` function, while Vite 8 has moved HMR/WebSocket configuration to `config.server.ws`.

The problematic code is at `defineClientValues` in the built dist:

```js
// dist/index.mjs around line 121
function defineClientValues(code, config) {
  let options = config.server.hmr;  // <-- DEPRECATED in Vite 8
  options = options && typeof options !== "boolean" ? options : {};
  const host = options.host || null;
  const protocol = options.protocol || null;
  const timeout = options.timeout || 3e4;
  const overlay = options.overlay !== false;
  let hmrPort;
  if (isObject(config.server.hmr)) {  // <-- DEPRECATED in Vite 8
    hmrPort = config.server.hmr.clientPort || config.server.hmr.port;
  }
  // ...
}
```

This should be migrated to use `config.server.ws` (the new Vite 8 API), with proper fallback logic to maintain backward compatibility.

### Reproduction

1. Create a Vite + Vue project with `@crxjs/vite-plugin`
2. Use Vite 8.x (e.g. `^8.0.12`)
3. Run `vite` or `vite dev`

Minimal config:

```ts
// vite.config.ts
import { crx } from '@crxjs/vite-plugin'
import vue from '@vitejs/plugin-vue'
import { defineConfig } from 'vite'
import manifest from './manifest.config.ts'

export default defineConfig({
  plugins: [
    vue(),
    crx({ manifest }),
  ],
  server: {
    port: 6886,
  },
})
```

**Expected**: No deprecation warning on startup

**Actual**: The deprecation warning is printed on every startup

### Logs

```shell
$ vite

  VITE v8.2.0  ready in 601 ms

  B R O W S E R
  E X T E N S I O N
  T O O L S

  ➜  CRXJS: Load dist as unpacked extension

  server.hmr.protocol/host/port/path/clientPort/timeout/server is deprecated. Use server.ws.* instead. Note that this option may be set by a plugin. Set VITE_DEPRECATION_TRACE=1 to see where it is called.
```

### System Info

```shell
System:
    OS: Windows 11 10.0.26200
    CPU: (x64)
  Binaries:
    Node: 24.17.0
    npm: 11.6.2
  npmPackages:
    @crxjs/vite-plugin: 2.7.1
    @vitejs/plugin-vue: 6.0.8
    vite: 8.2.0
```

### Related context

- Vite 8 migration guide: `server.hmr` → `server.ws` (https://vite.dev/config/server-options.html#server-ws)
- Similar issue in `vite-plugin-ruby`: https://github.com/ElMassimo/vite_ruby/issues/614

### Severity

annoyance

lKolabrodl:
Fixes #1231.

## Problem

Vite 8.1 moved WebSocket connection settings from `server.hmr` to `server.ws`. Starting a CRXJS development server with Vite 8.2.0 currently emits a deprecation warning on every run.

With `VITE_DEPRECATION_TRACE=1`, the warning points to CRXJS assigning `server.hmr.host`. Returning the complete `server` object from the config hook also makes Vite merge its compatibility proxy again, which triggers the same deprecated setters. The generated client values still read the old WebSocket fields as well.

## Fix

- Use `this.meta.viteVersion` from Vite's plugin context in the configuration and background-client load hooks to select the appropriate WebSocket settings.
- Return a narrow `server.ws` host override on Vite 8.1+, so the compatibility proxy is neither mutated nor merged back into the config.
- Keep `server.hmr` for Vite 3 through 8.0, including older versions without version metadata. Vite 8.0 only supports `server.ws: false`; connection options were added in 8.1.
- Read protocol, host, path, port, and timeout from the matching configuration while continuing to read `overlay` from `server.hmr`.
- Preserve disabled HMR/WebSocket settings.
- Cover version boundaries, absent or present `server.ws`, disabled settings, and metadata propagation through both plugin hooks with regression tests.

## Validation

- `pnpm exec vitest --run src/node --mode unit` — 93 tests passed.
- `pnpm exec vitest --run --mode compat` on the base Vite 3 dependency — 2 tests passed, 3 skipped by the existing suite.
- `pnpm run lint:types` — passed.
- `pnpm run lint:eslint` — passed; the final context type annotations were also checked with ESLint.
- `pnpm run lint:attw` — build and package type checks passed.
- Packed-plugin smoke checks with Vite 7.3.1, 8.0.0, 8.1.0, and 8.2.0 — development servers started, worker-client host/port/path/timeout values were verified, and no WebSocket deprecation warnings occurred with `VITE_DEPRECATION_TRACE=1`.

The full browser E2E matrix was not run locally.

### DIFF
--- a/.changeset/modern-ws-config.md
+++ b/.changeset/modern-ws-config.md
@@ -1,0 +1,5 @@
+---
+"@crxjs/vite-plugin": patch
+---
+
+Use Vite 8 WebSocket configuration without deprecated `server.hmr` fields while retaining compatibility with Vite 3 through 7.
--- a/packages/vite-plugin/src/node/defineClientValues.test.ts
+++ b/packages/vite-plugin/src/node/defineClientValues.test.ts
@@ -1,0 +1,116 @@
+import { describe, expect, it } from 'vitest'
+import { defineClientValues } from './defineClientValues'
+import type { ResolvedConfigWithHMRToken } from './types'
+
+const clientTemplate = [
+  'protocol=__HMR_PROTOCOL__',
+  'host=__HMR_HOSTNAME__',
+  'port=__HMR_PORT__',
+  'timeout=__HMR_TIMEOUT__',
+  'overlay=__HMR_ENABLE_OVERLAY__',
+].join('\n')
+
+function createConfig(server: Record<string, unknown>) {
+  return {
+    base: '/',
+    define: {},
+    mode: 'development',
+    server: {
+      middlewareMode: false,
+      port: 5173,
+      ...server,
+    },
+  } as unknown as ResolvedConfigWithHMRToken
+}
+
+describe('defineClientValues', () => {
+  it('reads WebSocket settings from Vite 8 without touching deprecated HMR fields', () => {
+    const hmr = { overlay: false }
+    for (const key of [
+      'clientPort',
+      'host',
+      'path',
+      'port',
+      'protocol',
+      'timeout',
+    ]) {
+      Object.defineProperty(hmr, key, {
+        get() {
+          throw new Error(`deprecated HMR field ${key} was read`)
+        },
+      })
+    }
+
+    const result = defineClientValues(
+      clientTemplate,
+      createConfig({
+        hmr,
+        ws: {
+          clientPort: 4321,
+          host: 'ws-host',
+          protocol: 'wss',
+          timeout: 1234,
+        },
+      }),
+    )
+
+    expect(result).toBe(
+      [
+        'protocol="wss"',
+        'host="ws-host"',
+        'port="4321"',
+        'timeout=1234',
+        'overlay=false',
+      ].join('\n'),
+    )
+  })
+
+  it('does not fall back to deprecated HMR fields when WebSockets are disabled', () => {
+    const hmr = { overlay: false }
+    Object.defineProperty(hmr, 'host', {
+      get() {
+        throw new Error('deprecated HMR host was read')
+      },
+    })
+
+    const result = defineClientValues(
+      clientTemplate,
+      createConfig({ hmr, ws: false }),
+    )
+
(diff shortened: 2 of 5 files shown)


FliPPeDround:
Thank you for your PR. Perhaps we could use Vite's [`plugin context meta`](https://vite.dev/guide/api-plugin#plugin-context-meta) API to distinguish between different versions and apply different HMR configurations accordingly. This would make the code more semantically clear. Would you be willing to update this PR? Much appreciated.

lKolabrodl:
Thanks for the suggestion, @FliPPeDround. Updated in a84eca1: both the config and background-client load hooks now use `this.meta.viteVersion` to select the WebSocket configuration.

I also checked the version boundary: connection options were added to `server.ws` in Vite 8.1, so Vite 3–8.0 continue using `server.hmr`. Older versions without metadata and disabled HMR/WebSocket settings remain supported.

Validation: 93 unit tests and the base compatibility suite passed, along with type, lint, build, and package checks. Smoke tests with Vite 7.3.1, 8.0.0, 8.1.0, and 8.2.0 confirmed server startup and worker-client values without WebSocket deprecation warnings. The full browser E2E matrix was not run locally.

PR #1240, opened 2026-09-05
chore: Migrate vite-plugin to tsdown and drop CommonJS support

FliPPeDround:
## Summary

This PR replaces Rollup with [tsdown](https://tsdown.dev) as the bundler for `@crxjs/vite-plugin` and makes the package ESM-only. The published output is functionally equivalent: same entry points, identical inlined client code, and an equivalent public type surface.

### Build system (Rollup → tsdown)

- Replaced `rollup.config.ts` (rollup 2 + esbuild plugin + rollup-plugin-dts) with a single `tsdown.config.ts` that emits `dist/index.mjs` (ESM) and `dist/index.d.mts` (declarations). A `build:done` hook strips an unused `import "acorn"` from the declarations — `acorn` is not a published dependency, so leaving it would break type resolution for strict consumers.
- The custom `bundleClientCode` plugin (inlines the `client/*` modules into the plugin as strings) was extracted to a shared `client-bundle-plugin.ts`, now used by both the tsdown build and the client-artifacts build in `tests/jest.globalSetup.ts` (rewritten accordingly, since it previously reused `rollup.config.ts`).
- `deps.neverBundle` replicates the old behavior of externalizing all dependencies, including devDependencies — e.g. the runtime `import('@vitejs/plugin-react')` in `plugin-contentScripts.ts` must not be bundled.
- Dependencies: removed `@rollup/plugin-*`, `rollup-plugin-dts`, `esbuild-runner` and `rimraf`. The runtime `rollup` dependency stays (the plugin calls the rollup API at runtime), as do `esbuild`/`rollup-plugin-esbuild` (dev-only, used for the client TS transform so the inlined client code stays byte-identical).

### ESM-only

- Removed `dist/index.cjs`, the root `index.cjs`/`index.d.cts` shims and `src/node/index.cjs.ts`.
- `exports` no longer has a `require` condition; `main`/`module`/`types` point at the ESM output.
- `lint:attw:check` now runs with `--ignore-rules cjs-resolves-to-esm`, the documented expected state for ESM-only packages.

### TypeScript 5

- Upgraded to TypeScript ^5.8.3 (required by tsdown / rolldown-plugin-dts) and `@types/node` ^20.
- Added `acorn` ^8.15 as a dev dependency: its own ESM-style declarations replace `@types/acorn`, whose CommonJS-style dts cannot be bundled by rolldown-plugin-dts.
- `client/*` imports now resolve through the ambient `declare module 'client/*'` in `modules.d.ts`; TS 5 no longer allows `.ts`-suffixed specifiers through tsconfig `paths` without `allowImportingTsExtensions`.
- `src/node/packageExports.test.ts` updated for the new exports shape.

## Verification

- `attw --pack` green for node10 / node16 (CJS) / node16 (ESM) / bundler
- `tsc --noEmit` and eslint pass
- Unit + output snapshot tests: 122 passed, no snapshot updates needed
- Compat tests pass on vite 3.2.11 / 6.4.1 / 7.3.6 / 8.1.5
- E2E tests pass with real Chromium (63 tests)
- Diffed the new `dist` output against the old Rollup output: all 9 inlined client strings are identical at runtime (8/9 byte-identical; the loading-page HTML differs only in `<\/script>` escaping when printed)

## Breaking change

`require('@crxjs/vite-plugin')` no longer works — use `import` (or dynamic `import()` from CommonJS). Covered by the `major` changeset.


packages/vite-plugin/package.json
```diff
@@ -27,43 +27,33 @@
...
-    "index.cjs",
-    "index.d.cts",
     "client.d.ts"
   ],
   "scripts": {
     "release": "bumpp",
     "prepublishOnly": "pnpm run build",
     "format": "prettier -w -c ../../.prettierrc.yaml",
-    "build": "run-s build:clean build:js",
-    "build:clean": "rimraf dist",
```
Toumash, line 60: why remove clean?

Toumash (approved):
✅ looks good, we can drop commonjs.

packages/vite-plugin/package.json
FliPPeDround: It was redundant: tsdown's built-in `clean: true` already wipes dist/ before each build, exactly what rimraf dist was doing for Rollup (which doesn't clean the outDir itself), so keeping the script and its deprecated rimraf dependency served no purpose.

PR #1242, opened 2026-09-07
fix(vite-plugin): close content script IIFE before sourcemap

### LINKED ISSUE #1220: IIFE for no-import content script closes after the sourcemap comment → "Unexpected end of input"
### Describe the bug

When a content script has **no imports/exports** (a self-contained module), `@crxjs/vite-plugin` wraps it in an IIFE by naive string concatenation, appending `})()` to code that already ends with rollup's `//# sourceMappingURL=…` line comment. Because that comment has no trailing newline, the appended `})()` lands *inside* the line comment and is stripped at parse time. The IIFE opened with `(function(){` never closes, and the script throws at runtime:

```
Uncaught SyntaxError: Unexpected end of input
```

Since the failing script runs at `document_start`, it never executes at all.

### Root cause

`packages/vite-plugin/…` (built output `dist/index.mjs:805` in v2.4.0):

```js
} else {
  bundleFileInfo.code = `(function(){${bundleFileInfo.code}})()
`;
}
```

At this point `bundleFileInfo.code` ends with:

```
…code…
//# sourceMappingURL=my-script.js.map
```

with **no trailing newline**, so the result is:

```js
(function(){…code…
//# sourceMappingURL=my-script.js.map})()   // <-- })() is now commented out
```

The `})()` is swallowed by the `//` comment, so the IIFE is never closed or invoked.

This only affects the `shouldUseLoader === false` branch (a chunk with zero `imports`, `dynamicImports`, and `exports`). Content scripts that import something get the loader path instead and are unaffected. It only manifests when `build.sourcemap` is enabled (the default in many setups), since without a sourcemap there's no trailing line comment to swallow the `})()`.

### Reproduction

1. Vite + `@crxjs/vite-plugin@2.4.0`, `build.sourcemap: true`.
2. Add a content script that imports nothing, e.g.:

   ```ts
   // early.ts — registered as a content_script in the manifest
   console.log('hello from document_start');
   ```

3. `vite build`.
4. Inspect the emitted chunk — the last line is:

   ```js
   …//# sourceMappingURL=early.ts-XXXX.js.map})()
   ```

5. Load the extension / run the file — `Uncaught SyntaxError: Unexpected end of input`.

### Expected behavior

The emitted chunk is valid JS: the `})()` closes and invokes the IIFE, and the sourcemap comment stays on its own line at the end, e.g.

```js
(function(){…code…})()
//# sourceMappingURL=early.ts-XXXX.js.map
```

### Suggested fix

Insert the closing `})()` before the trailing sourcemap comment rather than after it — e.g. split the sourcemap comment off, wrap the body, then re-append the comment. Using MagicString (which preserves the existing map) would be cleanest, but even a targeted string operation on the `//# sourceMappingURL=` suffix works.

### Workaround

A downstream `enforce: 'post'` `generateBundle` plugin that runs after crxjs and moves the stray `})()` back before the sourcemap comment:

```js
{
  name: 'fix-crx-iife-sourcemap',
  enforce: 'post',
  generateBundle(_options, bundle) {
    for (const file of Object.values(bundle)) {
      if (file.type !== 'chunk') continue;
      file.code = file.code.replace(
        /\n(\/\/# sourceMappingURL=[^\n]*?)\}\)\(\)\s*$/,
        '})()\n$1\n',
      );
    }
  },
}
```

### System Info

- `@crxjs/vite-plugin`: 2.4.0

dvd233:
## What changed

- keep trailing `sourceMappingURL` comments outside the generated content-script IIFE
- preserve LF/CRLF line endings, final-newline behavior, and the existing no-sourcemap output
- add syntax regression coverage and a patch changeset

## Why

For no-import content scripts, CRXJS adds scope isolation by appending `})()`.
Vite 8 can expose a trailing source-map line comment before CRXJS finalizes the
chunk. Appending the close/invocation after that comment comments it out and
emits JavaScript that fails with `Unexpected end of input`.

## Validation

- `pnpm --dir packages/vite-plugin exec vitest --run src/node/plugin-contentScripts.test.ts` — 4 passed
- `pnpm --dir packages/vite-plugin run test:run:out` — 157 passed, 1 skipped
- `pnpm --dir packages/vite-plugin run test:run:compat` — 2 passed, 3 skipped
- type lint, ESLint, Prettier, package build, and `attw` — passed
- Vite 8.1 A/B build: upstream output fails `node --check`; patched output passes

Fixes #1220


Toumash:
omg, perfect. Thank you @dvd233 for the contribution!

Toumash (approved):

PR #1244, opened 2026-09-12
fix(vite-plugin): discover inputs from build.rolldownOptions

Unheat:
## Summary

Teach the manifest plugin to discover additional entry points declared through Vite 8's `build.rolldownOptions.input`, while retaining compatibility with `build.rollupOptions.input`.

This prevents additional extension pages from being omitted from CRXJS's development dependency-discovery entries when a project adopts Vite 8's current configuration API.

## Background

Vite 8 uses Rolldown and documents `build.rolldownOptions` as the active configuration option. `build.rollupOptions` remains an alias but is deprecated:

- [Vite build options: `build.rolldownOptions`](https://vite.dev/config/build-options#build-rolldownoptions)
- [Vite build options: deprecated `build.rollupOptions`](https://vite.dev/config/build-options#build-rollupoptions)

CRXJS currently reads only `config.build.rollupOptions.input` from raw configuration inside the manifest plugin:

- [`plugin-manifest.ts` on `main`](https://github.com/crxjs/chrome-extension-tools/blob/main/packages/vite-plugin/src/node/plugin-manifest.ts#L109-L124)

Therefore, a Vite 8 configuration such as:

```ts
export default defineConfig({
  build: {
    rolldownOptions: {
      input: {
        offscreen: 'src/offscreen/offscreen.html',
        popup: 'popup.html',
      },
    },
  },
})
```

does not contribute those entries to the `optimizeDeps.entries` array returned by the CRXJS manifest config hook.

## Problem & Observed Behavior

In development mode, additional HTML entries declared only through `build.rolldownOptions.input` are absent from CRXJS's initial dependency discovery.

For an MV3 extension, this can cause Vite to discover and re-optimize dependencies on the fly after Chrome has already started loading extension contexts. If a service-worker module request receives Vite's `504 Outdated Optimize Dep` response during that process, Chrome rejects service worker registration with status code 3.

Using the deprecated `build.rollupOptions.input` alias or manually duplicating all entries in `optimizeDeps.entries` currently works around the problem, but CRXJS should accept Vite 8's replacement option natively.

## Changes

The manifest config hook now prefers `build.rolldownOptions.input` and falls back to `build.rollupOptions.input`:

```ts
const input =
  build?.rolldownOptions?.input ??
  build?.rollupOptions?.input ??
  []
```

This allows projects to use Vite 8's documented option without breaking compatibility with existing configurations and older supported Vite versions.

Existing behavior is fully preserved for:
- String, array, and object `input` shapes
- Absolute paths normalized relative to the Vite root
- Manifest-derived background, content-script, and HTML entries

## Verification & Tests

Added unit coverage in `plugin-manifest.test.ts` verifying that:
- Entries from `build.rolldownOptions.input` are merged into `optimizeDeps.entries` (Vite 8).
- Entries from `build.rollupOptions.input` remain supported for backward compatibility.

Local test run:
- `pnpm lint` -> ESLint & TypeScript passed
- `pnpm test:run:out` ->all test files passed, 1 skipped (159 tests passed, 1 skipped)


packages/vite-plugin/src/node/plugin-manifest.ts
```diff
@@ -109,7 +109,13 @@ export const pluginManifest: CrxPluginFn = () => {
           const { entries = [] } = config.optimizeDeps ?? {}
           // Vite ignores build inputs if optimize deps has explicit entries,
           // so we need to merge both to include extra HTML files
-          let { input = [] } = config.build?.rollupOptions ?? {}
+          const build = config.build as
+            | (typeof config.build & {
+                rolldownOptions?: NonNullable<typeof config.build>['rollupOptions']
```
Toumash: look good, but this type should be ideally on the original type somehow

